PR #4, opened 2023-07-04
Cmake install

gonzaponte:
Add instructions to CMake for the installation of nain4 as an independent library.


gonzaponte:
Replaced by #11.

PR #10, opened 2023-07-07
Add documentation on how the client can include Nain4 in their project

gonzaponte:
Describes how to include nain4 in client projects using CMake.

### DIFF
--- a/doc/src/how-to/enable-nain4-in-cmake.md
+++ b/doc/src/how-to/enable-nain4-in-cmake.md
@@ -11,16 +11,7 @@ There are 3 options to include nain4 in your project:
 Add this snippet to your CMakeLists.txt:
 
 ```
-include(FetchContent)
-
-FetchContent_Declare(
-  Nain4
-  GIT_REPOSITORY https://github.com/jacg/nain4.git
-  GIT_TAG        v0.0.0
-  SOURCE_SUBDIR  nain4
-)
-
-FetchContent_MakeAvailable(Nain4)
+{{#include ../../../client_side_tests/client_fetch_content/CMakeLists.txt:5:18}}
 ```
 
 The value given to `GIT_TAG` may be a replaced with a branch name (`origin/<branch-name>`) or a commit hash. When opting for the 'branch name' option, it's important to be aware of the potential risk of losing reproducibility. With each execution of CMake, the library may be updated to a different state, introducing changes that can impact the reproducibility of your project. This can lead to compilation or runtime errors, making it challenging to recreate specific build or runtime environments. Thus, we advise to use a commit hash or a tag instead.
@@ -31,7 +22,7 @@ The value given to `GIT_TAG` may be a replaced with a branch name (`origin/<bran
 Clone the Nain4 repository into your own project or create a symbolic link to the clone. Then simply add
 
 ```
-add_subdirectory(nain4/nain4)
+{{#include ../../../client_side_tests/client_subdirectory/CMakeLists.txt:5}}
 ```
 
 to your CMakeLists.txt.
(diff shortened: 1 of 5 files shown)


jacg:
Does this depend on #10?

gonzaponte:
It does. In the meantime, option 1 can be used if the user replaces the URL with one that points to my repo and the tag to my branch. Similarly, the other two options can be used if the client's clone of nain4 points to my branch.

jacg:
In that case, let's just merge, so that people can try it out without having to change anything else later on, besides the tag.

Can you prepare this for use with tag `0.0.0` in the main repo, please?

gonzaponte:
Just to be clear, as it stands it will work as soon as we merge #11. Is up to the client to decide whether to use a tag, a commit hash or origin/master. Do you want to set `GIT_TAG v0.0.0` in the documentation and clarify later that this can be replaced by a commit hash or origin/master?

jacg:
Ah, sorry, I got confused again about which of #10/#11 we're in.

I'll merge #11, and tag it with `v0.0.0`. Please adapt the docs PR (that'll be this one) so that the default instructions use that tag, and mention that this can be replaced with a commit etc.

We don't need to merge the #10 in any hurry, because I can publish the docs without merging, so we can iterate on the docs more freely.

doc/src/how-to/enable-nain4-in-cmake.md
```diff
@@ -1,3 +1,69 @@
...
+  Nain4
+  GIT_REPOSITORY https://github.com/jacg/nain4.git
+  GIT_TAG        v0.0.0
+  SOURCE_SUBDIR  nain4
+)
+
+FetchContent_MakeAvailable(Nain4)
+```
+
+The value given to `GIT_TAG` may be a replaced with a branch name (`origin/<branch-name>`) or a commit hash. When opting for the 'branch name' option, it's important to be aware of the potential risk of losing reproducibility. With each execution of CMake, the library may be updated to a different state, introducing changes that can impact the reproducibility of your project. This can lead to compilation or runtime errors, making it challenging to recreate specific build or runtime environments. Thus, we advise to use a commit hash or a tag instead.
```
jacg, line 26: Do we really want that `origin` in the branch name?

gonzaponte: Aha. This comes from when I tried to implement it with `ExternalProject`. The documentation said:

> Note that branch names and tags should generally be specified as remote names (i.e. origin/myBranch rather than simply myBranch). This ensures that if the remote end has its tag moved or branch rebased or history rewritten, the local clone will still be updated correctly.

However, the documentation for `FetchContent` doesn't say anything about this.

jacg:
With #13 merged, we should include the source code from the tests in these docs, and merge the docs source. Remember, the docs at https://jacg.github.io/nain4/ are published independently of `master`. Nevertheless, we do want the docs source to *be* on master, even if it is sometimes out of sync with the published version.

There is also the question of how to mitigate the need to bump tags manually in #13's tests and these docs.

jacg:
Tag `v0.0.1` is still valid, isn't it?

I don't really understand the `origin` business in the unresolved conversation.

gonzaponte:
> Tag v0.0.1 is still valid, isn't it?

I don't understand the question. v0.0.1 is valid, but I don't know what you are referring to.


> I don't really understand the origin business in the unresolved conversation.

IIUC, if you don't write `origin/` then FetchContent will only check `GIT_TAG` against the local clone. If something in the remote changes, the fetched content will not be affected. I guess that applies to tags as well, not only branch names...

jacg:
> I don't understand the question. v0.0.1 is valid, but I don't know what you are referring to.

We haven't merged anything that would require adding a new tag, and using it in the tests and docs.

> IIUC, if you don't write origin/ then FetchContent will only check GIT_TAG against the local clone. 

OK, but I would discourage using branch names: people should stick to immutable things like commit hashes (very immutable, very inconvenient, more fine-grained) and tags (not *guaranteed* to be immutable, but almost; convenient; less fine-grained control).

Then there's no guarantee that the remote will be called `origin`. OTOH, this clone is created by `FetchContent`? In which case the user would have to try really hard to make it be anything else.

In summary, we're happy for the `origin/` to be there in the docs?

doc/src/how-to/enable-nain4-in-cmake.md
```diff
@@ -1,3 +1,53 @@
...
+1. Adding it as an external project (recommended)
+2. Adding it as a subdirectory in your project
+3. Creating an independent installation
+
+## Adding Nain4 as a external project
+
+Add this snippet to your CMakeLists.txt:
+
+```
+{{#include ../../../client_side_tests/client_fetch_content/CMakeLists.txt:5:18}}
```
jacg, line 14: Ideally, we should use [anchors](https://rust-lang.github.io/mdBook/format/mdbook.html#including-portions-of-a-file) instead of line ranges. Not sure whether it's worth the effort. Your call.

gonzaponte:
> We haven't merged anything that would require adding a new tag, and using it in the tests and docs.

Nothing too relevant, I would say.

> but I would discourage using branch names

The documentation suggests using a tag. The option of using a branch is mentioned but discouraged for the same reasons you lay down here.

>  this clone is created by FetchContent?

This is how I understand it.


> In summary, we're happy for the origin/ to be there in the docs?

I don't have a strong feeling about this. The main reason that is there is because it was suggested by the cmake documentation, which favored being up to date with the remote tags. We are free to choose our own path.

doc/src/how-to/enable-nain4-in-cmake.md
gonzaponte: I like it. Will implement it.

jacg:
After stepping back to read it in context, I think it's spot on.

jacg:
I literally had my fingers on the keyboard to push the merge, when these last commits arrived? Is it ready now, or do you want to do any more?

gonzaponte:
I'm done

Issue #2, opened 2023-07-03
Provide an easy way to use `nain4`


jacg:
Implement whatever `cmake` incantations are necessary for `nain4` to be discoverable by `cmake` so that it can be used easily by clients.

NB: this repository contains
1. `nain4`
2.  a Nix flake for installing Geant4.
The latter is not needed to *use* the former: `nain4` needs *some* installation of Geant4, but not necessarily this one. People who already have Geant4 installed, will want to continue to use whatever installation they already had. The flake is merely a convenience for

1. Testing `nain4` on CI, etc.
2. People who haven't installed G4 yet.

jacg:
This is mostly covered by #10 and #11, but there are still issues with Catch2 on MacOS (and maybe elsewhere), as discussed in #15.

jacg:
I propose we close this, and use #15 to track the remaining issue we're aware of .

PR #11, opened 2023-07-07
Provide instructions to use Nain4 in other projects

gonzaponte:
Adds the necessary instructions to install and use nain4 in client projects.

#2

### DIFF
--- a/Nain4Config.cmake.in
+++ b/Nain4Config.cmake.in
@@ -1,0 +1,11 @@
+@PACKAGE_INIT@
+
+include("${CMAKE_CURRENT_LIST_DIR}/Nain4Targets.cmake")
+
+# Set variables for include directories and library dependencies
+set(NAIN4_INCLUDE_DIRS "@NAIN4_INCLUDE_DIRS@")
+set(NAIN4_LIB_DIRS "@NAIN4_LIB_DIRS@")
+set(NAIN4_LIBRARIES "@NAIN4_LIBRARIES@")
+
+include_directories(${NAIN4_INCLUDE_DIRS})
+link_directories(${NAIN4_LIB_DIRS})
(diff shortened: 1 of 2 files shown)


mdcge:
It seems that the names of this PR and #10 are the wrong way round and should be swapped.

gonzaponte:
Mmm, I didn't choose the best wording. This one refers to CMake commands on the Nain4-side in order to make the project more accessible from the client side. #10 contains the documentation on what the client should do to include Nain4 in their project. I will try to improve it, but suggestions are welcome.

jacg:
How can we test this automatically?

I know very little about `cmake`, so I'm confused about the various files:

+ `Nain4Config.cmake.in`
+ `Nain4Config.cmake`
+ `Nain4Targets.cmake`

and, above all, why two of them are not commited, even though `CMakeLists.txt` refers to them.

gonzaponte:
> How can we test this automatically?

#13 includes a test that verifies one of the clients' options to include nain4. The other options should also be tested, and I can add those tests provided you are happy with offering those solutions.

Regarding the files:

>  Nain4Config.cmake.in

This is a template file. There seems to be a convention where `.in` indicates that. CMake will replace several variables and provide

>    Nain4Config.cmake

in the installation directory, which will be used by the client to find nain4. This is why

>  two of them are not commited



>    Nain4Targets.cmake

is also generated by CMake. It is a bit of dark magic knowing that those files will be generated, but it is what it is.

On the other hand, I am not 100% convinced that all the changes I made are necessary for a successful integration, but I haven't tested it properly yet.

jacg:
As discussed [here](https://github.com/jacg/nain4/pull/10#issuecomment-1630391739), this was merged quickly to allow clients to easily follow the documentation (#10) that goes with it.

Issue #6, opened 2023-07-04
Warnings about shadowing when compiling with gcc


jacg:
When compiling `nain4` client code with gcc, it spews screenfuls of warnings about shadowing, drowning out any interesting warnings that might be emitted.

This needs to be silenced, without requiring the client to do anything. So, probably remove the shadowing from the `nain4` code, though I suspect it's perfectly kosher C++.

Or we could consider [gcc diagnostic pragmas](http://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html), but removing the shadowing is going to be a more general, more robust and less noisy approach.

gonzaponte:
Another warning to take care of, albeit of a different nature:
```
nain4.hh:66:16: warning: narrowing conversion of ‘std::forward<int>((* & args#0))’ from ‘int’ to ‘G4double’ {aka ‘double’} [-Wnarrowing]
   66 |   auto solid = new SOLID{std::forward<NAME>(name), std::forward<ArgTypes>(args)...};
```

gonzaponte:
Yes, it's not nain4's fault. However, it is slightly annoying to type the decimal point to avoid this warning. I was wondering if there is anything that can be done from this side.

jacg:
Yes, it's annoying. Just be glad it's not Rust, where it would be a hard error ;-)

I doubt that there's anything that can be done here, as it simply forwards whatever the client passes in, without any interference, to whatever underlying G4 constructor receives the information. It's a very general pattern, which allows a single implementation to cater for all the G4 solids, even ones which haven't been invented yet. That generality means that we can deal with any specifics like narrowing of integers to floats.

gonzaponte:
I'm working on this. Do we want to use the pragma directive for the whole file or every time there is a constructor that shadows a variable? The latter would introduce a lot of noise in the code, but it would be technically safer.

jacg:
I think that the cost of the extra noise outweighs the benefits of doing it properly. So let's just do it once per file.

Joe End User probably needs these warnings in user code, but we can reasonably expect a `nain4` *contributor* to be equipped to figure it out without the hint, if a shadowing problem arises in `n4` source code.

PR #12, opened 2023-07-07
Fix compilation warnings

gonzaponte:
Fixes compilation warnings in gcc. Some variables have been renamed to avoid shadowing.

One option to address #6.

### DIFF
--- a/nain4/nain4.hh
+++ b/nain4/nain4.hh
@@ -34,8 +34,8 @@
 
 
 // Disable GCC's shadow warnings
-// This happens in constructors where it is a C++ idiom to use
-// the same name for the argument and the class member.
+// This happens in constructors where it is a C++ idiom to have
+// constructor parameters match memeber names.
 // This has been reported to gcc's bug tracker
 // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78147
 #pragma GCC diagnostic push


jacg:
The semicolons in the first commit have to go, I completely agree.

As for the second commit, I fundamentally disagree with the compiler: This kind of shadowing *in constructors* is a common C++ idiom, and mangling the names to avoid the shadowing is a complete waste of time and space.

Looks like I'm not the only one: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78147. I completely agree with comment 7:

> I think that GCC should _never_ warn about this usage.  I really can't fathom a reason that anyone would want to enable shadow warnings in this situation

So, no, I'm not going to budge on this. It's a bug in the compiler. They should fix it.

In the meantime, what can we do to mitigate this deluge of noise? 

From the bug report, I infer that `gcc` doesn't distinguish between shadowing in constructors (where it is C++ idiom) and elsewhere (where it *might* cause confusion), so if we use [gcc diagnostic pragmas](https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html), they would have to be pushed and popped in a fairly fine-grained way. Nah, I don't want all that noise in our code. So I propose that we just disable the shadow warning once per file in *our* code.

gonzaponte:
I was also able to silent those warnings by removing the `-Wshadow` flag in `CMakeLists.txt` if you want to check this option. I didn't suggest it because I wasn't comfortable with disabling it _everywhere_. Do you want me to submit this option?

jacg:
You were right to be uncomfortable: we should not be disabling the shadowing warning everywhere; only in that particular usage in the constructors, where it has no business being on in the first place.

jacg:
Cachix binary cache seems to be working. I've populated it for Linux from my own machine, but we have to wait about an hour until the MacOS Geant4 is built on GHA and pushed to the cache. Then we can run the GHA tests with a turnaround of a few minutes, instead of the hour it used to take to recompile G4 each time. Then I can unleash all the blocked GHA runs on your many PRs.

nain4/nain4.hh
```diff
@@ -33,6 +33,14 @@
 #include <optional>
 
 
+// Disable GCC's shadow warnings
+// This happens in constructors where it is a C++ idiom to use
+// the same name for the argument and the class member.
```
jacg, line 38: Just a bit of my pet pedantry: **arguments** are *values* that are passed in to functions; the *names* to which those values are bound when they arrive inside the function are **parameters**.

About 90% of the time, we don't care about the difference, and we can conflate the two terminologies. But sometimes the difference *does* matter, and then it's useful to have the linguistic wherewithal to distinguish them, which is why I insist on using the correct terminology everywhere.

In this example you could even argue that your use of the term 'argument' is correct, because a parameter *is* a 'name for the argument', but let's just use 'parameter' anyway. 

TLDR: 'it is an idiom to have constructor parameters match member names'

gonzaponte: It's good that you keep pinpointing this because I keep forgetting :)

nain4/nain4.hh
```diff
@@ -34,8 +34,8 @@
 
 
 // Disable GCC's shadow warnings
-// This happens in constructors where it is a C++ idiom to use
-// the same name for the argument and the class member.
+// This happens in constructors where it is a C++ idiom to have
+// constructor parameters match memeber names'.
```
jacg, line 38: Stray quote/apostrophe at the end?

Nothing belongs to the names, so it's not possessive, and I can't see an opening quote.

gonzaponte:
Do you understand the build failure? Seems like it's not linking to G4.

jacg:
No, I don't understand it. But as I said in #15, we have to forget MacOS for a while, and make some progress on other things. It doesn't make sense to waste more time on MacOS, when we don't have access to a Mac and while tons of high-priority work is being delayed by it.

Issue #3, opened 2023-07-03
Re-enable use without flakes


jacg:
The `shell.nix` flake-compat shim was removed, because I thought nobody would need it, and that it was unnecessary noise. Turns out I was wrong.

jacg:
@gonzaponte could you please check whether this works for you?

gonzaponte:
Remind me, what should I do with this? I ran `nix-shell`, which took me to the configured shell, but I don't see a difference with the regular shell.

gonzaponte:
Hmm. I am pretty sure I had `just` and `clang` before. After I had run `nix develop` when you first told me to. Does this make sense?

jacg:
Yes. `nix develop` is, sort of, the new nix command equivalent of `nix-shell`.

You said that your OS was too old to have a modern Nix, so you couldn't use flakes. The `shell.nix` that was added here gives you (at least) two things:

1. `nix-shell` as an alternative for `nix develop`
2. The ability to use `direnv`.

gonzaponte:
`direnv` still tries to use the flake:
```
>>> direnv allow
direnv: loading ~/path/to/nain4/.envrc
direnv: using flake
environment:655: use_flake: command not found
```

jacg:
Which is hard to believe, given that in the [`.envrc`](https://github.com/jacg/nain4/blob/9004e598fd55a09eb8f2d6323282ca755d44d6e3/.envrc) that comes with [that commit](https://github.com/jacg/nain4/commit/9004e598fd55a09eb8f2d6323282ca755d44d6e3), there is no `use_flake`!

```
$ git show 9004e59:.envrc
if nix flake show &> /dev/null; then
  use flake
else
  use nix
fi
```

gonzaponte:
I almost forget about this.  Is there anything I can do to debug?
On a fresh shell, I do:
`setup nix`, which runs:
```
if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then
  . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
fi

eval "$(direnv hook bash)"
```
then `cd nain4` and I get the output of my previous comment.

jacg:
Well, first of all, show me the output of `cat .envrc`.

jacg:
And `direnv --version`.

gonzaponte:
As you say, there is no `use_flake`
```
if nix flake show &> /dev/null; then
  use flake
else
[...]

PR #13, opened 2023-07-07
Add client-side integration tests

gonzaponte:
Adds tests to verify the integration of the Nain4 project on the client side.


jacg:
Need to set up our own Nix [binary cache](https://nixos.wiki/wiki/Binary_Cache), to make CI tests run in a reasonable amount of time. Without the cache, it has to recompile Geant4 each time (unless we use the default configuration, which makes a multithreaded G4, which I think is a fundamentally bad idea).

I have one with [Cachix](https://www.cachix.org/) for rustpetalo, maybe nain4 can share that, for now.

gonzaponte:
Except for the recompilation test, I consider everything else up for review.

As for the recompilation test, I find that the environment is not fully preserved when running with `sudo -E`.

client_side_tests/justfile
```diff
@@ -0,0 +1,46 @@
+# -*-Makefile-*-
+
+# Run all tests in hand-written loop: more informative and colourful than ctest
```
jacg, line 3: There's no loop as such, any more, so this comment could confuse the reader.

.github/workflows/test.yml
```diff
@@ -41,7 +41,10 @@ jobs:
         run: nix develop
 
       - name: Test nain4
-        run: nix develop -c just
+        run: nix develop -c just test-nain4
```
jacg, line 44: The idea here was to check that the top-level justfile tests nain4 by default, so I don't really want this changed.

OTOH, it didn't really test that explicitly: you had to observe it in the GHA output, so maybe it's OK.

client_side_tests/client_fetch_content/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+cmake_minimum_required(VERSION 3.25)
+
+project(Client)
+
+include(FetchContent)
+
+FetchContent_Declare(
+  Nain4
+  GIT_REPOSITORY https://github.com/jacg/nain4.git
+  GIT_TAG        v0.0.0
```
jacg, line 10: The tag should be `v0.0.1` now.

gonzaponte:
> Can we import the tested CMakeLists.txt snippets into the documentation?

Sure, I will add them as examples

client_side_tests/client_independent_installation/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+project(Client)
+
+set(Nain4_DIR "$ENV{NAIN4_INSTALL}/lib/cmake/Nain4/")
+find_package(Nain4 REQUIRED)
+
+find_package(Geant4 REQUIRED)
+include(${Geant4_USE_FILE})
+
+add_executable(client_exe main.cxx)
```
jacg, line 11: `nain4` uses `.cc` and `.hh` as C++ filename extensions, so perhaps this should follow that pattern. Unless you deliberately want to distinguish client code from `nain4` code, by their extensions, which might be worthwhile.

client_side_tests/client_independent_installation/main.cxx
```diff
@@ -0,0 +1,14 @@
...
+  auto water = nain4::material("G4_WATER");
+  auto lx = 1.;
+  auto ly = 2.;
+  auto lz = 3.;
+
+  auto box = nain4::volume<G4Box>("test_box", water, lx, ly, lz);
+  nain4::place(box).at(0., 0., 0.).now();
+
+  return 0;
+}
```
jacg, line 14: Rather than copy-pasting the same source code in each test, how about softlinks?

.github/workflows/test.yml
gonzaponte: I thought that was the default behavior because there was no other case in the top just file. Now that I added another case I guessed that I needed to be more specific, but I didn't try without.

client_side_tests/justfile
```diff
@@ -0,0 +1,46 @@
...
+
+
+test-fetch-content:
+	#!/usr/bin/env bash
+	bash test_client_project_compiles.sh
+	if [[ $? -eq 0 ]]; then
+	  printf "\\033[32m => Client project compiled successfully with fetch content";
+	else
+	  printf "\\033[91m => Client project could not be compiled with fetch content"";
+	fi
```
jacg, line 19: This proto-test-framework in bash with the hard-wired ANSI escape codes already bothered me when I wrote them in the first place. Now that you've started copy-pasting it, it bothers me even more. But I guess we'll have to put up with it for now. I've been meaning to open an issue about it, but I can't think what to say.

.github/workflows/test.yml
jacg: The first recipe in a justfile is the default recipe, no matter how many there are.

jacg:
>>   Can we import the tested CMakeLists.txt snippets into the documentation?

> Sure, I will add them as examples

If possible, the sample that is already there should be included from the tests.

In general, wherever reasonably possible, code appearing in the docs should be automatically tested.

jacg:
Running `just test-client-side` locally, gives me the same error as on GHA:

```
just test-client-side
just client_side_tests/test-all
just test-fetch-content
bash: test_client_project_compiles.sh: No such file or directory
/tmp/just5tAdvs/test-fetch-content: line 18: unexpected EOF while looking for matching `"'
error: Recipe `test-fetch-content` failed with exit code 127
error: Recipe `test-all` failed on line 6 with exit code 127
error: Recipe `test-client-side` failed on line 7 with exit code 127
```
Looks like you have a double `"` in there.

client_side_tests/justfile
```diff
@@ -0,0 +1,46 @@
+# -*-Makefile-*-
+
+# Run all tests in hand-written loop: more informative and colourful than ctest
+
+test-all:
+  just test-fetch-content
+  just test-fetch-content-recompile
+  just test-subdirectory
+  just test-independent-installation
```
jacg, line 9: The success of this whole group depends only on the success of the last one.

jacg:
Another local failure that also appears on GHA:
```
$ just client_side_tests/test-fetch-content
bash: test_client_project_compiles.sh: No such file or directory
```

client_side_tests/client_fetch_content/CMakeLists.txt
gonzaponte: And it _needs to be_ `v0.0.1`

client_side_tests/client_independent_installation/CMakeLists.txt
gonzaponte: It was not intended, I'm just more use to that extension. Whatever you prefer.

client_side_tests/client_fetch_content/CMakeLists.txt
jacg: I think we're saying the same thing.

gonzaponte:
> If possible, the sample that is already there should be included from the tests

I don't understand this.


Regarding the failures, I think they have to do mostly with `just`. I don't have `just` installed, so I didn't test that part. Silly of me. I understand some failures so I will fix them.

client_side_tests/client_independent_installation/CMakeLists.txt
jacg: Let's stick to `.cc`/`.hh`.

And we'll also stick to snake case (and kebab case where syntactically possible) rather than Camel or StudlyCaps. I already find that this helps me understand what is `nain4` (and therefore subject to improvement) and what is imposed by G4 (and therefore we're stuck with it).

jacg:
>>  If possible, the sample that is already there should be included from the tests

> I don't understand this.

Stuff like [this](https://github.com/jacg/nain4/blob/c54392fd688ba7946370fd390ee8fa8f68e8358b/doc/src/how-to/enable-nain4-in-cmake.md?plain=1#L13-L24) should be included from the tests, rather than written by hand.

> I don't have `just` installed

`nix install nixpkgs#just` or `nix shell nixpkgs#just` or `nix run nixpkgs#just` ... ah but you have an ancient Nix ... so it's something like `nix-shell -p just` or some horror with `nix-env -i`. Having said that, your Nix probably supports the new `nix` command, but not flakes, so all the above might work.

It's also in the flake, which reminds me, I should do #3.

client_side_tests/justfile
gonzaponte: at the moment the tests just print something, but what would be the proper syntax? `&&`ing the output of each line?

jacg: > at the moment the tests just print something

Even so, two of them are failing, but the whole lot passes.

> &&ing the output of each line?

We'd need to `&&` the *statuses* of each line, making sure that all of them run. We don't want early failures to prevent the rest from running. I hope we can avoid repeating this sort of logic by hand, as I hacked together [here](https://github.com/jacg/nain4/blob/master/nain4/justfile#L9-L39)

jacg:
> Regarding nix. So far, I've been working in a non-nix environment. I should switch...

Not necessarily. Having a developer working in non-Nix, helps us spot things that don't work outside of Nix. Whether it's worth the hassle of working outside of Nix (and forgoing the convenience of always having a coherent set of dependencies, even when you switch commits), I don't know.

jacg:
> Stuff like [this](https://github.com/jacg/nain4/blob/c54392fd688ba7946370fd390ee8fa8f68e8358b/doc/src/how-to/enable-nain4-in-cmake.md?plain=1#L13-L24) should be included from the tests, rather than written by hand.

I've just noticed the tag in there. What can we do to avoid the need to have to update it by hand, every time a new tag is made?

jacg:
We may have to do it in the mdbook eventually, but right now, the tag needs to be set in the source code of the test, which (we agreed earlier) will be included by the book.

client_side_tests/justfile
```diff
@@ -0,0 +1,46 @@
...
+
+
+test-fetch-content:
+	#!/usr/bin/env bash
+	bash test_client_project_compiles.sh
```
jacg, line 14: `test_client_project_compiles.sh` does not exist (as mentioned elsewhere, but I'm mentioning it here again, as it's less likely to be overlooked here).

gonzaponte: Yes, I'm fixing these stupid errors. I used to run the tests myself with `bash filename.sh` and that way they passed, if you want to give it a go.

jacg: The chances of `bash test_client_project_compiles.sh` succeeding are pretty slim, if I don't have a copy of `test_client_project_compiles.sh`.

[Also, the independent installation (probably) doesn't make any sense on NixOS]

gonzaponte:
Ok, I got something more sensible now. I am not sure what's the right syntax to combine the results with `just`. `$var` doesn't seem to work.

jacg:
Write a [shebang recipe](https://just.systems/man/en/chapter_41.html), and you can use any language you like, including some shell language that understand `$var`.

client_side_tests/justfile
```diff
@@ -0,0 +1,46 @@
...
+  t1=0
+  t2=$(just test-subdirectory)
+  t3=$(just test-independent-installation)
+
+  [[ t0 && (t1 && (t2 && t3)) ]]
+
+
+test-fetch-content:
+  #!/usr/bin/env bash
+  bash test_client_project_compiles.sh
```
jacg, line 14: `test_client_project_compiles.sh` is *still* missing.

client_side_tests/justfile
```diff
@@ -0,0 +1,46 @@
+# -*-Makefile-*-
+
+test-all:
+  t0=$(just test-fetch-content)
+  t1=0
+  t2=$(just test-subdirectory)
+  t3=$(just test-independent-installation)
+
+  [[ t0 && (t1 && (t2 && t3)) ]]
```
jacg, line 9: Use a [shebang recipe](https://just.systems/man/en/chapter_42.html) to use a real shell, rather than `just`'s own language.

client_side_tests/justfile
gonzaponte: Agh, I must have messed up the rebase, apologies.

justfile
```diff
@@ -1,4 +1,11 @@
 # -*-Makefile-*-
 
+test-all:
+    just test-nain4
+    just test-client-side
```
jacg, line 5: The top-level `just` and `nain4/just` should **not** run the client side tests in their default recipe. I want `just` to be the thing you type to run your tests as you're developing, and the client side tests aren't helpful in that context. But they must run on CI. As it stands, they run 3 times on CI:

1. Test nain4
2. nix-shell
3. Run client-side tests

If we remove them from `test-all`, the `test-all` name becomes a misnomer. Perhaps make `test-nain4` the default recipe. Though I'm not sure of the value of a `test-all` that includes the client side tests, as they should run as a separate step on CI.

Anyway, the client-side tests seem to pass on CI on Linux. Still the same error on MacOS.

jacg:
(Sorry, posted this in the wrong PR, earlier.)

Have you any idea where the "C compiler identification" messages come from?

https://github.com/jacg/nain4/actions/runs/5529534966/jobs/10087735944#step:8:199

Or why this one linked above should generate a syntax error?

client_side_tests/justfile
```diff
@@ -0,0 +1,47 @@
...
+
+
+test-fetch-content:
+  #!/usr/bin/env bash
+  bash test_fetch_content.sh
+  if [[ $? -eq 0 ]]; then
+    printf "\\033[32m => Client project compiled successfully with fetch content\n";
+  else
+    printf "\\033[91m => Client project could not be compiled with fetch content\n";
+  fi
```
jacg, line 20: This pattern is repeated 4 times, so it needs abstracting.

Just recipes are essentially functions, so how about passing the status code and the text that varies, as arguments to a recipe that implements the colouring and the yea-or-nay part of the text?

gonzaponte:
> Have you any idea where the "C compiler identification" messages come from?

Do you mean which cmake line? I have no idea.

> Or why this one linked above should generate a syntax error?

I believe the syntax error is not related to that. Seems like there is some text being overwritten and partially hiding the error. I'm trying to reproduce locally, but I'm not able to.

gonzaponte:
Fixed the syntax error.

jacg:
Our force-pushes are interfering. I removed the experiment with reordering `target_link_libraries` (which was motivated by [this](https://github.com/jacg/nain4/issues/15#issuecomment-1632169896) BTW.) You put it back along with your useful commits. Heads up: I've just removed it again and force-pushed. (And I'll do any other experiments on another branch.)

client_side_tests/justfile
jacg: The print recipe addresses as much of this as is reasonable to do in this PR.

client_side_tests/justfile
jacg: If you're happy with the solution you have to this, please mark this as resolved.

client_side_tests/justfile
```diff
@@ -0,0 +1,43 @@
+# -*-Makefile-*-
+
+test-all:
+  #!/usr/bin/env sh
+  just test-fetch-content           ;(( status |= $? ))
+  $(exit 0)                         ;(( status |= $? ))
+  just test-subdirectory            ;(( status |= $? ))
+  just test-independent-installation;(( status |= $? ))
```
jacg, line 8: Mustn't forget the disabled test here.

jacg: So, is `fetch-content-recompile` is the only thing left to do in this whole PR?

gonzaponte: It is. I haven't found a solution yet.

jacg: In that case, I'm tempted to merge, and open an issue about it.

gonzaponte: Your call.

jacg: You open the issue, I'll merge. Anything you want to do before I do that?

gonzaponte: The skipped test is included in the PR. If you are happy leaving it there, I have nothing else to do here.

gonzaponte:
Mmm, I had tested it locally and it worked as it was supposed to... I'm on it.

PR #16, opened 2023-07-11
Set default value for CMAKE_INSTALL_PREFIX only if not given by CLI

gonzaponte:
In #11 the installation path was forced inadvertently. Here we allow changing this path using the CLI like `-DCMAKE_INSTALL_PREFIX=/path/to/install/`


jacg:
Is there anything sensible I can do with this (without getting too distracted from what I was doing) other than merging it and tagging it with `v0.0.1`?

PR #21, opened 2023-07-12
Use Catch2 v3

### LINKED ISSUE #18: Consider upgrading Catch2 to v3
v2 was header-only, v3 provides libraries. Recompiling tests' main function takes an eternity, and I had to come up with a hack that minimizes the need to do this. But it still takes a significant portion of the time GHA actions take to run.

But I'm not at all sure whether it would be worth the effort, and what downsides there might be.

It might help with some teething troubles on MacOS, but we shouldn't put too much weight on that, as in the long term, Catch2 (like all dependencies) will be provided robustly and automatically by Nix (maybe with the help of some sort of container), so this will be a non-issue ... eventually.

gonzaponte:
A preview of the changes needed to update Catch2 to v3 (which resolves #18). The interface doesn't change much and the compilation times improve significantly.
However, running `just` seems to trigger the compilation of `nain4-tests` even though it is already built.

If accepted, do we add it to the flake or rely on `FetchContent`?

### DIFF
--- a/nain4/CMakeLists.txt
+++ b/nain4/CMakeLists.txt
@@ -57,19 +57,7 @@ set(Nain4_TESTS
 )
 
 # ----- Use Catch2 as C++ testing framework ---------------------------------
-set(FETCHCONTENT_UPDATES_DISCONNECTED ON
-    CACHE BOOL
-    "Cache package contents to avoid unnecessary downloads")
-
-include(FetchContent)
-FetchContent_Declare(
-  Catch2
-  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
-  GIT_TAG        v3.3.2
-  OVERRIDE_FIND_PACKAGE
-)
-FetchContent_MakeAvailable(Catch2)
-
+find_package(Catch2 REQUIRED)
 set(ALL_TEST_SOURCES
   test/catch2-main-test.cc
   # ${Nain4_SOURCES}
@@ -96,7 +84,6 @@ target_include_directories(
   $<INSTALL_INTERFACE:include>
 )
 
-list(APPEND CMAKE_MODULE_PATH ${catch2_SOURCE_DIR}/extras)
 include(CTest)
 include(Catch)
 catch_discover_tests(nain4-tests)
--- a/nain4/CMakeLists.txt
+++ b/nain4/CMakeLists.txt
@@ -57,7 +57,19 @@ set(Nain4_TESTS
 )
 
 # ----- Use Catch2 as C++ testing framework ---------------------------------
-find_package(Catch2 REQUIRED)
+set(FETCHCONTENT_UPDATES_DISCONNECTED ON
+    CACHE BOOL
+    "Cache package contents to avoid unnecessary downloads")
+
+include(FetchContent)
+FetchContent_Declare(
+  Catch2
+  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+  GIT_TAG        v3.3.2
+  OVERRIDE_FIND_PACKAGE
+)
+FetchContent_MakeAvailable(Catch2)
+
 set(ALL_TEST_SOURCES
   test/catch2-main-test.cc
   # ${Nain4_SOURCES}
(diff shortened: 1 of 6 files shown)


jacg:
> If accepted, do we add it to the flake or rely on `FetchContent`?

The flake is the canonical source of truth about dependencies. Anything else is extras we provide for the convenience of poor souls who cannot (or refuse to) use Nix. Eventually we'll figure out how to build Apple silicon containers from Nix, and then, hopefully, all the other cruft will become obsolete. Don't hold your breath, but we'll get there eventually.

So please adapt `flake.nix` accordingly. You can find the relevant package name [here](https://search.nixos.org/packages?channel=23.05&from=0&size=50&sort=relevance&type=packages&query=catch2).

Now, the whole point of https://github.com/jacg/nain4/blob/v0.0.1/nain4/test/catch2-main-test.cc was something to do with reducing the compile times, by getting the other tests to use it (or maybe its build-products?), but I forget the details. It perturbs me a little that no tests related to this break. I'll have to look into it.

nain4/CMakeLists.txt
```diff
@@ -57,7 +57,19 @@ set(Nain4_TESTS
...
+    "Cache package contents to avoid unnecessary downloads")
+
+include(FetchContent)
+FetchContent_Declare(
+  Catch2
+  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+  GIT_TAG        v3.3.2
+  OVERRIDE_FIND_PACKAGE
+)
+FetchContent_MakeAvailable(Catch2)
```
jacg, line 71: Let the flake take care of this.

nain4/test/test-nain4.cc
```diff
@@ -17,7 +17,11 @@
 #include <G4Material.hh>
 #include <G4Gamma.hh>
 
-#include <catch2/catch.hpp>
+#include <catch2/catch_approx.hpp>
+#include <catch2/generators/catch_generators.hpp>
+#include <catch2/catch_test_macros.hpp>
+
+using namespace Catch;
```
jacg, line 24: I'm no fan of `using namespace ...;` How about `namespace c2 { using namespace Catch; }` ?

OTOH, I guess that in `v2` we already had something that was equivalent to `using namespace Catch;`.

Still, if they created the namespace, perhaps we should take advantage of it. Perhaps they added it because there's more stuff in there in `v3` with concomitant greater risk of trouble. So, try the above trick, and see how much noisier the client code gets.

gonzaponte: This was mostly to avoid `Catch::Approx` so we can bring into the scope that function

nain4/CMakeLists.txt
gonzaponte: I was hoping this would aid solve the MacOS problems.

nain4/test/test-nain4.cc
jacg: In that case it should be `using Catch::Approx;`

nain4/CMakeLists.txt
jacg: All these ad-hoc hacks are only short-term (add-on) solutions. If we try to maintain them, it will be an absolute nightmare and endless drain of resources. The correct solution is to get it to work in Nix. Probably 'all' that is required there, is to fix whatever problems there are in the Geant4 derivations for Darwin. Unfortunately, as I don't have a Mac, and I have blissfully not cared about the evolution of Apple's idiosyncratic build mechanisms for about 16 years, I am not the best person to attempt this, and there are more urgent things to be done.

So, I don't expect to get there any time soon, but we *will* get there, and then the problem will be solved once and for all. Getting things set up in Nix is a huge pain in the arse, but once you get there, the benefits are enormous. It's difficult to fathom just how enormous, for people who have not experienced the results.

jacg:
> It perturbs me a little that no tests related to this break.

Somehow, in spite of having looked for it, I failed to notice that you modified this file.

So, this all LGTM ... if we're convinced that v3 is a step in the right direction. I haven't looked into v3 at all, yet.

jacg:
I've had a quick look at the `v3` docs, and I don't see any reason not to migrate. So I'm going to merge this, unless I hear something to the contrary, soonish.

jacg:
I've reverted the merge, because some client-side tests fail (and CI fails to report that they fail).

jacg:
Good, CI now shows the [failure](https://github.com/jacg/nain4/actions/runs/5544770074/jobs/10122719394).

You deleted your branch, so I had to push it to my own.

PR #22, opened 2023-07-13
Add boolean volumes

### LINKED ISSUE #5: G4SubtractionSolid
Hi!

Is there a way to perform a subtraction between two solids with nain4? Or is it necessary to do it with Geant4 and then define a G4LogicalVolume and do G4PVPlacement?

Thanks!

gonzaponte:
A simple proposal to close #5.

Pro: extremely simple
Con: creates unnecessary logical volumes

IMHO, the pro outweighs the con.

### DIFF
--- a/nain4/test/test-nain4.cc
+++ b/nain4/test/test-nain4.cc
@@ -27,6 +27,8 @@
 
 #include <numeric>
 
+using std::make_tuple;
+
 TEST_CASE("nain material", "[nain][material]") {
 
   // nain4::material finds the same materials as the verbose G4 style
@@ -174,6 +176,46 @@ TEST_CASE("nain volume", "[nain][volume]") {
   CHECK(solid->GetCubicVolume() / m3 == Approx(8 *  lx    * ly    * lz     / m3));
   CHECK(solid->GetSurfaceArea() / m2 == Approx(8 * (lx*ly + ly*lz + lz*lx) / m2));
   CHECK(solid->GetName()             == "test_box");
+}
+
+TEST_CASE("nain boolean", "[nain][boolean]") {
+  auto water = nain4::material("G4_WATER");
+  auto l     = 2 * m;
+  auto dr    = G4ThreeVector{l/2, l/2, l/2};
+  auto norot = nullptr;
+  auto vbox  = l*l*l;
+  auto vint  =     vbox / 8.;
+  auto vsum  = 2 * vbox - vint;
+  auto vsub  =     vbox - vint;
+
+  auto box1 = nain4::volume<G4Box>("test_box1",      water, l/2, l/2, l/2);
+  auto box2 = nain4::volume<G4Box>("test_box2", NOMATERIAL, l/2, l/2, l/2);
+
+  auto union1 = nain4::boolean<ADD      >("test_add"      , box1, box2, norot, dr);
+  auto union2 = nain4::boolean<JOIN     >("test_join"     , box1, box2, norot, dr);
+  auto subtr1 = nain4::boolean<SUB      >("test_sub"      , box1, box2, norot, dr);
+  auto subtr2 = nain4::boolean<SUBTRACT >("test_subtract" , box1, box2, norot, dr);
+  auto inter1 = nain4::boolean<INT      >("test_int"      , box1, box2, norot, dr);
+  auto inter2 = nain4::boolean<INTERSECT>("test_intersect", box1, box2, norot, dr);
+
+  //                        object  volume name
+  auto items = { make_tuple(union1, vsum, "add")
+               , make_tuple(union2, vsum, "join")
+               , make_tuple(subtr1, vsub, "sub")
+               , make_tuple(subtr2, vsub, "subtract")
+               , make_tuple(inter1, vint, "int")
+               , make_tuple(inter2, vint, "intersect")
+               };
+
+  auto density = water->GetDensity();
+  for (auto& item : items){
+    auto [obj, vol, name] = item;
+    CHECK(obj->TotalVolumeEntities() == 1);
+    CHECK(obj->GetMass() / kg        == Approx(vol * density / kg));
+    CHECK(obj->GetMaterial()         == water);
+    CHECK(obj->GetName()             == std::string("test_") + name);
+  }
+
 }
 
 TEST_CASE("nain place", "[nain][place]") {
--- a/nain4/test/test-nain4.cc
+++ b/nain4/test/test-nain4.cc
@@ -60,7 +60,7 @@ TEST_CASE("nain material", "[nain][material]") {
     auto name = "n4test_FR4";
     auto density = 1.85 * g/cm3;
     auto state = kStateSolid;
-    auto [nH, nC, nO] = std::make_tuple(12, 18, 3);
+    auto [nH, nC, nO] = make_tuple(12, 18, 3);
 
     // Make the material using nain4::material_from_elements
     auto fr4 = nain4::material_from_elements_N(name, density, state,
@@ -110,7 +110,7 @@ TEST_CASE("nain material", "[nain][material]") {
     auto name = "n4test_LYSO";
     auto density = 7.1 * g/cm3;
     auto state = kStateSolid;
-    auto [fLu, fY, fSi, fO] = std::make_tuple(0.714, 0.040, 0.064, 0.182);
+    auto [fLu, fY, fSi, fO] = make_tuple(0.714, 0.040, 0.064, 0.182);
 
     // Make the material using nain4::material_from_elements
     auto lyso = nain4::material_from_elements_F(name, density, state,
--- a/nain4/nain4.hh
+++ b/nain4/nain4.hh
@@ -4,6 +4,7 @@
 #define nain4_hh
 
 #include <CLHEP/Geometry/Transform3D.h>
+#include <G4IntersectionSolid.hh>
 #include <G4LogicalVolume.hh>
 #include <G4LogicalVolumeStore.hh>
 #include <G4Material.hh>
@@ -18,8 +19,10 @@
 #include <G4String.hh>
 #include <G4SolidStore.hh>
 #include <G4SDManager.hh>
+#include <G4SubtractionSolid.hh>
 #include <G4ThreeVector.hh>
 #include <G4Transform3D.hh>
+#include <G4UnionSolid.hh>
 #include <G4VPhysicalVolume.hh>
 #include <G4VSensitiveDetector.hh>
 #include <G4VUserActionInitialization.hh>
@@ -74,6 +77,25 @@ G4LogicalVolume* volume(NAME name, G4Material* material, ArgTypes&&... args) {
   auto solid = new SOLID{std::forward<NAME>(name), std::forward<ArgTypes>(args)...};
   return new G4LogicalVolume{solid, material, solid->GetName()};
 }
+
+#define NOMATERIAL nullptr
+
+#define ADD       G4UnionSolid
+#define JOIN      G4UnionSolid
+#define SUB       G4SubtractionSolid
+#define SUBTRACT  G4SubtractionSolid
+#define INT       G4IntersectionSolid
+#define INTERSECT G4IntersectionSolid
+
+template<class BOOL, class NAME, class... ArgTypes>
+G4LogicalVolume* boolean(NAME name, G4LogicalVolume* main, G4LogicalVolume* secondary, ArgTypes&&... args) {
+  return volume<BOOL>( name
+                     , main     ->GetMaterial()
+                     , main     ->GetSolid()
+                     , secondary->GetSolid()
+                     , std::forward<ArgTypes>(args)...);
+}
+
 
 // --------------------------------------------------------------------------------
 // Utilies for concisely retrieving things from stores


jacg:
Be a bit more creative in your use of [issue closure keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

How about changing your first line to
> A simple proposal to close #5.

?

Similarly, chuck in a keyword [here](https://github.com/jacg/nain4/pull/21#issue-1801428424).

As much as I love all the work you are doing here, and the rate at which you are doing it, reviewing it all is preventing me from making much progress on writing the documentation that this project desperately needs. So, could you please add one or two relevant pages to the docs, that describe this solution, directly in this PR?

Recall that we have four categories of documentation, and something like this would need something in at least *How-To* and *Reference*, and maybe even in *Tutorial*. Writing the reference is likely to be a bit tedious, and it won't be stable until we've finalized the details, so that can be left until later.

Ideally, we want to see the plain G4 version next to the n4 one, to help us evaluate whether it is worth the effort.

Similarly for #14, which I haven't had the time to look at, yet.

In general, we shouldn't add any new utilities without documentation ... and tests, but I know that I don't need to tell you about the tests.

gonzaponte:
> Be a bit more creative in your use of [issue closure keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

I'm aware of this. I tend to add them later on when the PR is more mature, but I'm ok with doing it from the beginning.

> could you please add one or two relevant pages to the docs?

Sure thing.

> Similarly for https://github.com/jacg/nain4/pull/14, which I haven't had the time to look at, yet.

Don't waste your time on that. It's not the most urgent feature and it's not working.

nain4/test/test-nain4.cc
```diff
@@ -176,6 +178,46 @@ TEST_CASE("nain volume", "[nain][volume]") {
...
+
+  auto box1 = nain4::volume<G4Box>("test_box1",      water, l/2, l/2, l/2);
+  auto box2 = nain4::volume<G4Box>("test_box2", NOMATERIAL, l/2, l/2, l/2);
+
+  auto union1 = nain4::boolean<ADD      >("test_add"      , box1, box2, norot, dr);
+  auto union2 = nain4::boolean<JOIN     >("test_join"     , box1, box2, norot, dr);
+  auto subtr1 = nain4::boolean<SUB      >("test_sub"      , box1, box2, norot, dr);
+  auto subtr2 = nain4::boolean<SUBTRACT >("test_subtract" , box1, box2, norot, dr);
+  auto inter1 = nain4::boolean<INT      >("test_int"      , box1, box2, norot, dr);
+  auto inter2 = nain4::boolean<INTERSECT>("test_intersect", box1, box2, norot, dr);
```
jacg, line 199: IIUC, your solution always combines two volumes to produce a third. What happens if I want to use that in a further boolean operation?

In the very vague thoughts I had about this when opening #5, I was thinking of an interface that would allow chaining, or even arbitrary trees of operations.

As I don't recall having used these boolean operations myself in real code, I have no clue whether real Geant4 use cases have a need for this, even though it would be the right the to do from an idealistic perspective.

nain4/test/test-nain4.cc
```diff
@@ -176,6 +178,46 @@ TEST_CASE("nain volume", "[nain][volume]") {
...
+  auto items = { make_tuple(union1, vsum, "add")
+               , make_tuple(union2, vsum, "join")
+               , make_tuple(subtr1, vsub, "sub")
+               , make_tuple(subtr2, vsub, "subtract")
+               , make_tuple(inter1, vint, "int")
+               , make_tuple(inter2, vint, "intersect")
+               };
+
+  auto density = water->GetDensity();
+  for (auto& item : items){
```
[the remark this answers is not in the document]
gonzaponte: I did, but the keywords `ADD`, etc. are class names and I couldn't make it work.

nain4/test/test-nain4.cc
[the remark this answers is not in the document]
gonzaponte: In the constructors for boolean solids there are only two options: rotation + translation or transform. `nain4::boolean` just forwards the parameters. Do you have any suggestions?

jacg:
>  I tend to add them later on when the PR is more mature

Given that the point of the PR is to close that issue, I'd say it should go in from the beginning.

nain4/test/test-nain4.cc
gonzaponte: > What happens if I want to use that in a further boolean operation?

I haven't tried, but I wouldn't expect a problem with that. Maybe I'm wrong.

jacg: I'm not expecting any technical problems, just verbosity. So the question is whether we can come up with a better interface for chaining boolean operations.

gonzaponte: I think we can. But I have the feeling that the final interface can be decoupled from this implementation. We can offer something similar to `place` that uses this functionality behind the scenes.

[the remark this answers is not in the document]
jacg: A builder which, when told to build, inspects the internal state it has accumulated, and forwards it to the G4 constructor that is appropriate. Or the most general one, filling in the blanks with null operations.

Looking at [this](https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/geometry/solids/Boolean/include/G4SubtractionSolid.hh#L50-L63) it looks like you've overlooked the rotation matrix?

Also, looking at the 2nd and 3rd constructor in there, it looks like there are mutually exclusive alternatives: transform-3d vs 3-vec and rot-matrix. So the builder would have to prevent you from mixing them, at compile time. 

More generally, at some stage I have to take a step back and see if there's a *bearable* way of writing typestate patterns in C++. I really want to promote the multitude of G4 pitfalls to compile-time errors.

jacg: Well, yes, this would be an implementation details of the interface. So the client would end up not seeing this at all.

We also need to have a good look at [`G4MultiUnion`](https://geant4-forum.web.cern.ch/t/creating-a-g4unionsolid-with-10-different-components/9906/6).

We have to be wary of creating beautiful, convenient interfaces, which generate inefficient code and promote bad practices.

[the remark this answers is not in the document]
gonzaponte: > A builder which, when told to build, inspects the internal state it has accumulated, and forwards it to the G4 constructor that is appropriate. Or the most general one, filling in the blanks with null operations.

I understand. I will have to try it and see.

> Looking at [this](https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/geometry/solids/Boolean/include/G4SubtractionSolid.hh#L50-L63) it looks like you've overlooked the rotation matrix?

I don't follow. The `norot` variable in the test is there precisely because I need to pass it as an argument.

gonzaponte: > We also need to have a good look at [G4MultiUnion](https://geant4-forum.web.cern.ch/t/creating-a-g4unionsolid-with-10-different-components/9906/6).

I'm aware of this, but I postponed its implementation to a later stage since it needs special care.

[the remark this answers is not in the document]
jacg: > The `norot' variable in the test is there precisely because I need to pass it as an argument.

Sorry, brain fart.

jacg:
> Con: creates unnecessary logical volumes

I guess there's no reason for this, other than you reusing `n4::volume`?

Given that all this boolean solid stuff uses `G4VSolids` as inputs and outputs, there really shouldn't be any logical volumes at all.

I'm thinking about the more general interface discussed in one of the review conversations, and I don't see any reason why there should be any logical volumes.

jacg:
What would be the problems with an interface like this:
```c++
    G4VSolid* first = ...;
    // etc.
    n4::boolvol(first)
        .add(second).at(position).rot(rotaion)
        .sub(third).trans(transform_3d)
        .name("Opitional override of default name constructed from components and ops")
        .now();
```
?

gonzaponte:
> I guess there's no reason for this, other than you reusing n4::volume?

Correct.

> What would be the problems with an interface like this

I don't see a problem. But I do appreciate skipping the somewhat unnecessary step of creating a solid that adds more noise than signal.

jacg:
And it would subsume the use cases from your tests, e.g.

```c++
auto union1 = n4::boolean<ADD>("test_add", box1, box2, norot, dr);
```
would become
```c++
auto union1 = n4::boolean(box1).add(box).at(dr).now();
auto onion1 = n4::boolean(box1).add(box).at(dr).name("test_add").now(); // If you *really* want the name
```

jacg:
>  But I do appreciate skipping the somewhat unnecessary step of creating a solid that adds more noise than signal.

Indeed. So `.now()` can be bifurcated into two methods, one which returns a solid, one which gives you the logical volume. Ideas for pithy but sufficiently clear names?

jacg:
> Ideas for pithy but sufficiently clear names?

+ `{solid,logical}`
+ `get_{solid,logical}`
+ `make_{solid,logical}`

gonzaponte:
I do like the syntax though. I think it is very clear and self-explanatory. The main inconvenience I find at the moment is that the input solids need to be created explicitly. So we either add an interface in n4 (which doesn't add anything over G4?) or we let the user create them in G4 (which looks dirty to me).

> So .now() can be bifurcated into two methods

Is there a case for returning solids? Or is it just for flexibility?

gonzaponte:
>> Is there a case for returning solids? 

> Depends on whether we find a nice interface for constricting more complex boolean hierarchies. With .solid() it just works:

Yes, I agree. It can help the assembly of complex boolean compositions.

> How about ...

I like it. In practice, the case of pre-constructed solids will probably prevail, but that's as much as can be done, I think.

jacg:
> Depends on whether we find a nice interface for constricting more complex boolean hierarchies. With `.solid()` it just works

Actually, `.add()` and friends could accept an `n4::boolean`, so we wouldn't need `.solid()` for that use case. But still, it's trivial, and adds a bit of flexibility at very little cost.

gonzaponte:
So to summarize, we are going directly to this interface, right?

jacg:
Ooops, the proposed interface doesn't mention materials, which are needed if we want it to return logical volumes.

Maybe a `.material()` method, which turns the product from solid to logical? But then you would need to ensure that it's given at each step ... preferably at compile time ... which would be a tricky typestate to code in C++.

jacg:
Ah, it's OK: the whole thing must have the same material, right?. So 

+ `.solid()`
+ `.logical(material)`

would be enough.

We could even do 

+ `.now()` for solid
+ `.now(material)` for logical

but that's probably too cute and cryptic.

jacg:
> The templated methods you proposed can become difficult to read relatively easily, I think.

You mean because of the bazillion arguments that they take, they would drown out the structure with low-level noise? Yes, quite possibly. So do you think we should *not* provide them?

> So we either add an interface in n4 (which doesn't add anything over G4?)

I do think that 
```c++
n4::boolean(thys).add(that).at(position).logical(water);
```
is a huge improvement over ... well, I can't be bothered to figure out the equivalent bare G4.

[Aside: looking at the above, it might be nice if the end were spelt `.material(water)`, but perhaps that doesn't make it sufficiently obvious that a logical volume is being made at this point.]

> or we let the user create them in G4 (which looks dirty to me).

I don't think there is much that can be done to improve the situation: these solids do need a bunch of different construction arguments, and they have to be passed in *somewhere*. If all you're doing is creating a simple solid, then I don't think n4 can improve over G4 in any *significant* way.

So I think it's perfectly OK for the earlier example to be preceded with 

```c++
auto thys = new G4Sphere(lots, of, noisy, arguments, that, must, be, given, somewhere);
```

jacg:
> I don't think n4 can provide anything better than G4, which annoys me a bit.

Te-he. Me too. Just a little bit. But this *is* an example of where G4 doesn't do anything terribly wrong. Probably because it's difficult to imagine how to mess up something so simple.

[Hmm, now you've got me thinking about whether these solids' multiple parameters can be organized more conveniently.]

> And mixing G4 with n4 in the client code feels dirty.

I know what you mean, but we have to be pragmatic and not excessively purist. n4 tries to help you write G4 code, where it can: it's not a complete replacement for G4. I think it's an important message to send to potential users, that n4 does not force you into anything: the whole of G4 (familiarity, warts, bugs and all) remains available to you, but we can help to make *some* things easier, clearer and more robust.

gonzaponte:
What's the reason for `child` to be optional in `n4::place`?

jacg:
> What's the reason for `child` to be optional in `n4::place`?

Hmm. Two things, I guess

+ There is no way of preventing the client from passing in a `nullptr`
+ You might want to specify *where* things will be placed, earlier, and later on say *what* should be placed there. 
   ```c++
   auto put_somewhere_special = n4::place().in(mother).at(position).rot(rotation);
   auto my_precious_thing = ... // complex logic
   put_somewhere_special.child(my_precious_thing).now();
   ```
   Pay careful attention to where `.now()` is written (and, perhaps more importantly, where it is *not*).
   Of course, for this to work, we'd need a `.child()` method (with some good name), and there isn't one, yet.

More generally, `nullptr` doesn't distinguish between 

1. This hasn't been specified
2. This has been specified to be NULL

and `option` does.

So all our, as yet unimplemented, typestates should give compile-time errors if you try to build before all of the mandatory values have been set.

In other words, it's probably there, setting us up for the conceptually correct thing to do, but a number of things have not been implemented in this case, yet.

Does this make any sense?

gonzaponte:
Yes. I was slightly confused because I didn't see any reason for a childless case, mostly. So if IIUC, *at the moment* optional's job is to protect against nullptrs?

jacg:
No, I don't think it protects against `nullptr`s at the moment. The constructor accepts them, and `.now()` doesn't perform any checks.

gonzaponte:
No, but it gives you an error, instead of a segfault, which is a small improvement.

PR #24, opened 2023-07-13
Use Catch2 v3 properly

gonzaponte:
Does what #21 did + addresses missed test failures.

One client-side test was failing. The issue was that `FetchContent` was coded to download tag `v0.0.1`, which uses Catch2 v2. However, the environment (which is set by the flake) contains Catch v3, which is not compatible. 

I've added a commit that forces the test to download this branch (with Catch2 v3). After demonstrating that this works, I propose we merge this pr, tag it with some version, and commit another change to the test file that points to this new tag.

### DIFF
--- a/client_side_tests/client_fetch_content/CMakeLists.txt
+++ b/client_side_tests/client_fetch_content/CMakeLists.txt
@@ -9,8 +9,8 @@ set(FETCHCONTENT_UPDATES_DISCONNECTED ON CACHE BOOL "Cache package contents to a
 
 FetchContent_Declare(
   Nain4
-  GIT_REPOSITORY https://github.com/jacg/nain4.git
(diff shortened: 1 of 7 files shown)


jacg:
It works. 

Shall I merge, as it is?

gonzaponte:
Yes

jacg:
Merged and tagged with `v0.0.2`.

gonzaponte:
you take care of the commit or should I open a PR?

jacg:
I did it. It didn't work. What have I done wrong?

jacg:
Ah, it's still using your fork.

jacg:
It seems to have worked now.

I guess the tag should be moved up, in case someone checks it out and runs the client-side tests.

PR #28, opened 2023-07-14
Compile Geant4 with debug symbols

### LINKED ISSUE #27: Compile Geant4 with debugging symbols
Segfaults are a fact of life in Geant4. Without debug symbols, their cause is an even greater mystery.

I think I managed this before, with an overlay like this
```nix
geant4-debug-overlay = final: previous: {
   geant4 = previous.geant4.overrideAttrs (old: {
     cmakeBuildType = "Debug";
   });
};
```
but my initial attempts appear not to succeed, by which I mean that `gdb $G4_DIR/lib/libG4FR.so` claims that 
> (No debugging symbols found in /nix/store/fjwwybxxxlckp523fr152jz8pw994lc3-geant4-11.0.4/lib/libG4FR.so)

jacg:
A, thus far, failed attempt to fix #27.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -9,18 +9,30 @@
   };
 
   outputs = { self, nixpkgs, flake-utils, flake-compat }:
+    let
+      geant4-dbg-overlay = final: prev: {
+        geant4-dbg = prev.geant4.overrideAttrs (old: {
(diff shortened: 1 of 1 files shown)


jacg:
This is now being done elsewhere.

Issue #7, opened 2023-07-05
Provide default nain4 app and/or template


jacg:
EDIT: this (see task list, below) needs to be split into separate, smaller issues, which can be knocked off in reasonable amounts of time.

Provide a default app which allows a new user to write a fully-working Geant4 application in only a few lines of code.

It should be possible to use this as a starting point for writing your own applications, by gradually replacing the defaults with your own logic.

Ideally, it would also come with automated tests (and a GitHub Action), in order to encourage the user to write and run tests from the very beginning.

EDIT: we are making a nix flake template for bootstrapping client nain4 projects
TASK LIST
+ [ ] Add instructions to README
+ [ ] Add documentation to the mdbook
+ [ ] Add tests for the template
+ [ ] Add GHA steps which catch template-breaking nain4 changes
+ [ ] Add exhaustive CHANGEMEs
+ [ ] Add more functionalities to the app
+ [ ] Configure binary cache with `nixConfig` in the template's flake. (This whole topic appears to be almost completely undocumented in Nix, but here's an example in the wild: https://github.com/divnix/digga/blob/main/examples/devos/flake.nix#L4-L14)

gonzaponte:
I understand that we want two things:
1. a working example
2. a template to be filled (without replacements)

is this correct?

As for 1. IIUC we want a main that can discover the geometry, actions, etc. defined by the user and run them. In that way, the user would only need to compile their code linking it to the app?

And for 2. we want a simple main that uses empty instances (geometry, actions, etc) in the same file that the user needs to fill, right?

gonzaponte:
Ok, I think I understand.

jacg:
Also, for the documentation to be digestible, the code in the tutorials and examples **cannot** be spread out over 13 files, and really shouldn't waste space on defining class boilerplate in the same file, à la [trivial-full-app-test.cc](https://github.com/jacg/nain4/blob/master/nain4/test/trivial-full-app-test.cc). In fact, we want something like that, but without the classes.

We need to find an app that strikes a good balance between being not too trivial on the one hand, and not too long on the other.

And we probably need to do #8 first. Ideally the user clones the repo, types `just start` and gets a pretty Qt UI showing a geometry ... and can proceed from there.

gonzaponte:
> In fact, we want something like that, but without the classes.

Seeing that test got me walking in the wrong direction, but I'm on a good track now.

jacg:
[...]

Issue #8, opened 2023-07-05
Provide an abstraction for initialization of the user interface.


jacg:
Essentially, hide the following boilerplate behind some interface:
```c++
// Initialize visualization
std::unique_ptr<G4VisManager> visManager = std::make_unique<G4VisExecutive>();
// G4VisExecutive can take a verbosity argument - see /vis/verbose guidance.
// G4VisManager* visManager = new G4VisExecutive("Quiet");
visManager -> Initialize();

// Get the pointer to the User Interface manager
G4UImanager* UImanager = G4UImanager::GetUIpointer();
std::unique_ptr<G4UIExecutive> ui;
if ( argc == 1 ) { ui = std::make_unique<G4UIExecutive>(argc, argv); }
if ( ! ui ) {
    // batch mode
    G4String command = "/control/execute ";
    G4String fileName = argv[1];
    UImanager->ApplyCommand(command+fileName);
} else {
    // interactive mode
    UImanager->ApplyCommand("/control/execute init_vis.mac");
    ui->SessionStart();
}
```

jacg:
Done in 1428d00.

PR #29, opened 2023-07-14
Add n4app

gonzaponte:
A first go at #7. This includes an app that runs the most basic simulation imaginable (or close to). The app comes with its own `CMakeLists` and `justfile` for convenience. I've also added the corresponding documentation.

Notice that `nain4` has to be installed for the example to link against it. It can also be done with `FetchContent`, but I think the former option makes more sense in this context. Let me know if you think otherwise.

### DIFF
--- a/doc/src/how-to/build-a-minimal-nain4-app.md
+++ b/doc/src/how-to/build-a-minimal-nain4-app.md
@@ -17,17 +17,20 @@ In order to run a simulation we need to create a run manager:
 {{#include ../../../examples/n4app.cc:create_run_manager}}
 ```
 
-The run manager needs to be initialized with 4 attributes:
+The run manager needs to be initialized with 3 attributes:
 - A physics list
 - A geometry
-- A primary generator
 - A set of actions
 
+To build the action set we need to provide a primary generator action. Other actions are optional.
 Note that the physics list must be instantiated and given to the run manager before the primary generator is instantiated. We provide the run manager with these attributes:
+
 ```c++
 {{#include ../../../examples/n4app.cc:build_minimal_framework}}
 ```
+
 here, `my_geometry`, and `my_generator` are the two functions we need to provide to run our simulation:
+
 ```c++
 {{#include ../../../examples/n4app.cc:my_geometry}}
 
--- a/doc/src/how-to/build-a-minimal-nain4-app.md
+++ b/doc/src/how-to/build-a-minimal-nain4-app.md
@@ -2,18 +2,18 @@
 
 The most basic example of a nain4 app can be found in [examples/n4app.cc](../../../examples/n4app.cc). Here is the full file:
 
-```
+```c++
 {{#include ../../../examples/n4app.cc:full_file}}
 ```
 
 Now, let's split it into bite-sized portions. First we create a simple program that reads the number of events from CLI:
-```
+```c++
 {{#include ../../../examples/n4app.cc:pick_cli_arguments}}
 {{#include ../../../examples/n4app.cc:closing_bracket}}
 ```
 
 In order to run a simulation we need to create a run manager:
-```
+```c++
 {{#include ../../../examples/n4app.cc:create_run_manager}}
 ```
 
@@ -24,11 +24,11 @@ The run manager needs to be initialized with 4 attributes:
 - A set of actions
 
 Note that the physics list must be instantiated and given to the run manager before the primary generator is instantiated. We provide the run manager with these attributes:
-```
+```c++
 {{#include ../../../examples/n4app.cc:build_minimal_framework}}
 ```
 here, `my_physics_list`, `my_geometry`, and `my_generator` are the three functions we need to provide to run our simulation:
-```
+```c++
 {{#include ../../../examples/n4app.cc:my_physics_list}}
 
 {{#include ../../../examples/n4app.cc:my_geometry}}
@@ -37,16 +37,16 @@ here, `my_physics_list`, `my_geometry`, and `my_generator` are the three functio
 ```
 
 With this, our setup is ready, and we can start the simulation with
-```
+```c++
 {{#include ../../../examples/n4app.cc:run}}
 ```
 
 Don't forget to add the relevant includes
-```
+```c++
 {{#include ../../../examples/n4app.cc:includes}}
 ```
 
 and deleting the run_manager before exiting for proper memory deallocation.
-```
+```c++
 {{#include ../../../examples/n4app.cc:close_gracefully}}
 ```
--- a/doc/src/how-to/build-a-minimal-nain4-app.md
+++ b/doc/src/how-to/build-a-minimal-nain4-app.md
@@ -27,10 +27,8 @@ Note that the physics list must be instantiated and given to the run manager bef
 ```c++
 {{#include ../../../examples/n4app.cc:build_minimal_framework}}
 ```
-here, `my_physics_list`, `my_geometry`, and `my_generator` are the three functions we need to provide to run our simulation:
+here, `my_geometry`, and `my_generator` are the two functions we need to provide to run our simulation:
 ```c++
-{{#include ../../../examples/n4app.cc:my_physics_list}}
-
 {{#include ../../../examples/n4app.cc:my_geometry}}
 
 {{#include ../../../examples/n4app.cc:my_generator}}
@@ -46,7 +44,4 @@ Don't forget to add the relevant includes
 {{#include ../../../examples/n4app.cc:includes}}
 ```
 
-and deleting the run_manager before exiting for proper memory deallocation.
-```c++
-{{#include ../../../examples/n4app.cc:close_gracefully}}
-```
+To run this test type `just test-examples` from the top `nain4` directory.
--- a/doc/src/how-to/build-a-minimal-nain4-app.md
+++ b/doc/src/how-to/build-a-minimal-nain4-app.md
@@ -1,0 +1,52 @@
+# How to create a minimal nain4 app
+
+The most basic example of a nain4 app can be found in [examples/n4app.cc](../../../examples/n4app.cc). Here is the full file:
+
+```
+{{#include ../../../examples/n4app.cc:full_file}}
+```
+
+Now, let's split it into bite-sized portions. First we create a simple program that reads the number of events from CLI:
+```
+{{#include ../../../examples/n4app.cc:pick_cli_arguments}}
+{{#include ../../../examples/n4app.cc:closing_bracket}}
+```
+
+In order to run a simulation we need to create a run manager:
+```
+{{#include ../../../examples/n4app.cc:create_run_manager}}
+```
+
+The run manager needs to be initialized with 4 attributes:
+- A physics list
+- A geometry
+- A primary generator
+- A set of actions
+
+Note that the physics list must be instantiated and given to the run manager before the primary generator is instantiated. We provide the run manager with these attributes:
+```
+{{#include ../../../examples/n4app.cc:build_minimal_framework}}
+```
+here, `my_physics_list`, `my_geometry`, and `my_generator` are the three functions we need to provide to run our simulation:
+```
+{{#include ../../../examples/n4app.cc:my_physics_list}}
+
+{{#include ../../../examples/n4app.cc:my_geometry}}
+
+{{#include ../../../examples/n4app.cc:my_generator}}
+```
+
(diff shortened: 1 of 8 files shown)


doc/src/how-to/build-a-minimal-nain4-app.md
```diff
@@ -0,0 +1,52 @@
+# How to create a minimal nain4 app
+
+The most basic example of a nain4 app can be found in [examples/n4app.cc](../../../examples/n4app.cc). Here is the full file:
+
+```
```
jacg, line 5: Need to tell it's c++ code:
```
```c++
```
here and in all the other includes.

examples/n4app.cc
```diff
@@ -0,0 +1,92 @@
...
+  // ANCHOR: run
+  run_manager -> BeamOn(n_events);
+  // ANCHOR_END: run
+
+  // ANCHOR: close_gracefully
+  delete run_manager;
+  // ANCHOR_END: close_gracefully
+
+// ANCHOR: closing_bracket
+  return 0;
```
jacg, line 89: Ugh!

That `0` is a magic number. If you insist on writing this line, then use `EXIT_SUCCESS` (that way it will continue to work even if the code is ever deployed on an OS where success is indicated with something other than `0`).

But you shouldn't waste space on writing that line at all, because the C++ standard says that `EXIT_SUCCESS` is returned implicitly from `main` if nothing is returned explicitly.

It's a shame that most IDE's write this crap for you when you open a new `.cc` file ... along with the truly egregious `using namespace std;`. At least I haven't seen `void main` for a while.

examples/n4app.cc
```diff
@@ -0,0 +1,92 @@
...
+  run_manager -> SetUserInitialization(actions);
+  run_manager -> Initialize();
+  // ANCHOR_END: build_minimal_framework
+
+  // ANCHOR: run
+  run_manager -> BeamOn(n_events);
+  // ANCHOR_END: run
+
+  // ANCHOR: close_gracefully
+  delete run_manager;
```
jacg, line 85: Nooooo! We do not want to see `delete` *anywhere* (except low-level library code) in modern C++!

Use smart pointers. Usually `std::unique_ptr`.

We don't want to see `new` either, but that's more difficult to avoid completely, so the rule of thumb is, you are only allowed to write `new` as an argument to something that will take responsibility of managing the resource you just created).

Geant4 makes it *impossible* to follow this last rule, so we have to tolerate `new`s in other places too, but only where it's completely unavoidable.

examples/n4app.cc
```diff
@@ -0,0 +1,92 @@
...
+#include "g4-mandatory.hh"
+
+#include <G4RunManagerFactory.hh>
+#include <G4SystemOfUnits.hh>
+#include <G4VPhysicalVolume.hh>
+#include <G4Event.hh>
+#include <G4Box.hh>
+#include <FTFP_BERT.hh>
+
+#include <stdlib.h>
```
jacg, line 13: Please don't include legacy `<*.h>`s in C++, unless you time travel to before 1997 (or thereabouts).

Anything that the C standard library provides as `<thing.h>` you should use in C++ as `<cthing>`. It's essentially the same, but with namespaces.

examples/n4app.cc
```diff
@@ -0,0 +1,92 @@
...
+}
+// ANCHOR_END: my_geometry
+
+
+// ANCHOR: my_generator
+void my_generator(G4Event* event) {
+  auto geantino = n4::find_particle("geantino");
+  auto vertex   = new G4PrimaryVertex();
+  vertex->SetPrimary(new G4PrimaryParticle(geantino, 1, 0, 0));
+  event->AddPrimaryVertex(vertex);
```
jacg, line 41: I'm partial to whitespace around `->`: it helps humans to parse the text. Especially in long chains, which are pretty common in G4.

examples/n4app.cc
```diff
@@ -0,0 +1,92 @@
...
+}
+// ANCHOR_END: my_generator
+
+
+// ANCHOR: my_physics_list
+G4VUserPhysicsList* my_physics_list() {
+  return new FTFP_BERT{0};
+}
```
jacg, line 49: This could be written on 1 line, instead of 3.

examples/n4app.cc
```diff
@@ -0,0 +1,92 @@
...
+// ANCHOR_END: my_physics_list
+
+// ANCHOR: pick_cli_arguments
+int main(int argc, char* argv[]) {
+  if (argc != 2) {
+    std::cerr << "Wrong number of parameters: " << argc << std::endl;
+    print_usage();
+    return 1;
```
jacg, line 57: `return EXIT_FAILURE`.

examples/n4app.cc
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  auto n_events = atoi(argv[1]);
```
jacg, line 60: `std::stoi` please.

jacg (commented):
First batch of feedback, while I dig into the details of further comments, some of which will take some time to get right.

jacg:
`just examples/` gives me an error complaining about `CMakeLists.txt`. That's an out-of-the-box experience that will need improving.

jacg:
> (I haven't included the app test in the GHA yet)

Any reason not to?

It certainly should go there ASAP, so that things like `just examples/` not working is clearly signalled to us.

examples/n4app.cc
```diff
@@ -0,0 +1,92 @@
...
+  auto* run_manager = G4RunManagerFactory::CreateRunManager(G4RunManagerType::Default);
+  // ANCHOR_END: create_run_manager
+
+  // ANCHOR: build_minimal_framework
+  // Important! physics list has to be set before the generator!
+  run_manager -> SetUserInitialization(my_physics_list());
+
+  auto geometry  = new n4::geometry {my_geometry};
+  auto generator = new n4::generator{my_generator};
+  auto actions   = new n4::actions  {generator};
```
jacg, line 73: This violates the rule that `new` is only allowed as an argument to something that takes responsibility for the lifetime. So it should be something like

```c++
  run_manager -> SetUserInitialization(new n4::geometry {my_geometry});
  run_manager -> SetUserInitialization(new n4::actions{new n4::generator{my_generator}});
  run_manager -> Initialize();
```
Which is, perhaps, a bit ugly, but it sets us up for the chained use of the actions' set methods (which will require a set of parentheses around the `new n4::actions ...`), at which point it becomes less ugly than standard G4 style.

We should probably create some n4 functions that absolve the user from typing `new`, and which will probably add some safety in the process. And they'll get rid of the need for the extra parentheses before using the setters.

I'll open an issue for this.

jacg (commented):
Here are a few more.

examples/n4app.cc
```diff
@@ -0,0 +1,92 @@
...
+  auto n_events = atoi(argv[1]);
+  // ANCHOR_END: pick_cli_arguments
+
+  // ANCHOR: create_run_manager
+  auto* run_manager = G4RunManagerFactory::CreateRunManager(G4RunManagerType::Default);
+  // ANCHOR_END: create_run_manager
+
+  // ANCHOR: build_minimal_framework
+  // Important! physics list has to be set before the generator!
+  run_manager -> SetUserInitialization(my_physics_list());
```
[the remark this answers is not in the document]
jacg: Though the run-time error G4 issues in this case is surprisingly good, IIRC.

We must verify the exact constraints and make sure we say something that is exactly, not approximately, true ... as far as reasonably possible.

In the docs you say that

> the physics list must be instantiated and given to the run manager before the primary generator is instantiated.

*Exactly* what  must precede *exactly* what else? Is it the instantiations, or telling the manager about them, or one of each?

But really, we should just make this something that the user simply doesn't have to be aware of.

gonzaponte:
> just examples/ gives me an error complaining about CMakeLists.txt. That's an out-of-the-box experience that will need improving.

Hmm, it worked (and works) locally. I'll let the GHA tell me what's wrong.

> Any reason not to?

I forgot and I had to go offline for a while.

> Actually, I'd be tempted to write it on zero lines: just write
> run_manager -> SetUserInitialization(FTFP_BERT{0});
> later on.

For some reason, I cannot reply to this inline. My logic here was that it would be more obvious that these are the components that the user has to provide. Hiding them in separate functions feels more didactic to me. It can also ease the transition to building a more complex application.

The same goes for inlining the creation of `n4::geometry` and such. I know conciseness is a virtue, but how about
```
auto geometry = std::make_unique<n4::geometry>{my_geometry};
...
run_manager -> SetUserInitialization(geometry.release())
```

jacg (commented):
A few more.

Overall, I like it a lot.

Slight shame that you can't launch a graphical viewer and see pretty pictures. Perhaps we can do #8, and then adding the viewer will we a question of only 2 or 3 more lines. But that can be done in a later PR.

jacg:
> > Actually, I'd be tempted to write it on zero lines: just write
> > run_manager -> SetUserInitialization(FTFP_BERT{0});
> > later on.
> 
> For some reason, I cannot reply to this inline. My logic here was that it would be more obvious that these are the components that the user has to provide. Hiding them in separate functions feels more didactic to me. 

Yes, as I said, it does make the code more self-documenting.

> how about
> 
> ```
> auto geometry = std::make_unique<n4::geometry>{my_geometry};
> ...
> run_manager -> SetUserInitialization(geometry.release())
> ```

I'd really rather not confuse newbies with `make_unique` and `.release()`

(BTW: `make_unique` is a function, so it takes `()` not `{}`)

examples/n4app.cc
```diff
@@ -0,0 +1,92 @@
...
+    std::cerr << "Wrong number of parameters: " << argc << std::endl;
+    print_usage();
+    return 1;
+  }
+
+  auto n_events = atoi(argv[1]);
+  // ANCHOR_END: pick_cli_arguments
+
+  // ANCHOR: create_run_manager
+  auto* run_manager = G4RunManagerFactory::CreateRunManager(G4RunManagerType::Default);
```
jacg, line 64: We should probably explicitly ask for the `SerialOnly` run manager.

gonzaponte: This begs for the creation of `std::unique_ptr<G4RunManager> n4::run_manager()`

doc/src/how-to/build-a-minimal-nain4-app.md
```diff
@@ -0,0 +1,52 @@
...
+In order to run a simulation we need to create a run manager:
+```
+{{#include ../../../examples/n4app.cc:create_run_manager}}
+```
+
+The run manager needs to be initialized with 4 attributes:
+- A physics list
+- A geometry
+- A primary generator
+- A set of actions
```
jacg, line 24: This is not *quite* true. Though this might be a case where simplicity is better than truth.

It needs 3 things

+ physics list
+ geometry
+ actions

and the actions have one mandatory component, the generator, and a bunch of optional ones.

(which explains the structure of `n4::actions`)

examples/n4app.cc
jacg: Indeed. But only as a prelude to #31.

jacg:
When we're close to done (which is pretty soon, probably) we should get our users (both of them!) to read it, and give some feedback.

jacg:
I added `n4::run_manager()` and rebased on top of it. Then I shortened the code a bit more, and realized it was close to fitting on a single screen, which is always very useful for didactic examples (and elsewhere), so I reduced the length a bit more.

Also added header discovery in clangd. 

As you're unavailable, I'll declare this first iteration done, and merge.

We can always improve it later, revert my changes if you don't like them, etc.

jacg:
I'm not entirely convinced that this is a How-To and not a tutorial, but we can ponder that later, in the context of other docs that we create.

jacg:
The `CMakeLists.txt` wasn't mentioned at all. That's an oversight.

jacg:
@Hiitsaso, @mdcge could you please read [this](https://jacg.github.io/nain4/how-to/build-a-minimal-nain4-app.html) and give us your feedback.

gonzaponte:
> Also added header discovery in clangd.

Having to add that snippet every time is somewhat annoying. Specifically, it's easy to forget because I don't need it. I don't understand why, do you? I am also using the nix environment (or so I think), so it should be the same for everyone.

I'm happy with your changes.

jacg:
It's for `clangd` which is an LSP server. If you don't use LSP in your editor, then you won't see the difference.

Without it, LSP can't find the headers, and is essentially useless.

PR #43, opened 2023-07-29
Use macros to avoid repetition of `n4::solid` members that store and set `x`, `y`, `z`, `r`, `phi`, `theta` etc. 

jacg:
If you're happy with this, go ahead and merge.

### DIFF
--- a/nain4/n4-volumes.hh
+++ b/nain4/n4-volumes.hh
@@ -29,6 +29,7 @@ enum class BOOL_OP { ADD, SUB, INT };
 
 struct boolean_shape;
 
+// ---- Base class for interfaces to G4VSolids --------------------------------------------------------
 struct shape {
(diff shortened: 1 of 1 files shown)


jacg:
`G4Cons` uses 2 `r`s, and `G4Tubs` uses 2 `x`s and 2 `y`s.

I've added an `N` parameter to the macros, to allow them to be reused for these cases, but I don't think that it should go in this PR.

gonzaponte (approved):
Makes a lot of sense to avoid repetitions.

- G4Cons can also be implemented using only these macros if we allow `HAS_R` to add a suffix to the `r*` variables. Something like `TYPE& rEXTRA_inner` etc.
- With this implementing the `xy` etc. methods should be trivial

But in any case that can go in a future PR. I'm merging this.

PR #55, opened 2023-07-31
Make n4::ui pass on integer argument to /run/beamOn

jacg:
This allows a third CLI use case for `n4::ui`:

1. `executable                # runs GUI`
2. `executable macs/run.mac   # executes run.mac in batch mode`
3. `executable 42             # executes /run/beamOn 42 in batch mode`

I thought we had an issue for this, but I don't see it.


jacg:
... which would just execute `/run/beamOn N` after the macro file has finished?

nain4/n4_ui.cc
```diff
@@ -6,23 +6,55 @@
...
+    auto file_name =                argv[1] ;
+    if (n.has_value()) { beam_on  (n        ); }
+    else               { run_macro(file_name); }
+  }
+  if (argc == 3) {
+    auto n         = parse_unsigned(argv[2]);
+    auto file_name =                argv[1] ;
+    if (n.has_value()) {
+      run_macro(file_name);
+      beam_on  (n        );
```
gonzaponte, line 56: Have you checked what happens with `beamOn(0)`?

nain4/n4_ui.cc
```diff
@@ -6,23 +6,55 @@
...
+  // 1 argument on CLI: run in batch mode
+  //   +     integer: <ARGUMENT-FOR-BEAM-ON>
+  //   + non-integer: <MACRO-FILE>
+  if (argc == 2) {
+    auto n         = parse_unsigned(argv[1]);
+    auto file_name =                argv[1] ;
+    if (n.has_value()) { beam_on  (n        ); }
+    else               { run_macro(file_name); }
+  }
+  if (argc == 3) {
```
gonzaponte, line 51: Document `argc == 3` like the other two branches

nain4/n4_ui.cc
jacg: Nothing exciting: 
```
./n4app-00-basic 0

          ################################
          !!! G4Backtrace is activated !!!
          ################################


**************************************************************
 Geant4 version Name: geant4-11-00-patch-04    (3-March-2023)
                       Copyright : Geant4 Collaboration
                      References : NIM A 506 (2003), 250-303
                                 : IEEE-TNS 53 (2006), 270-278
                                 : NIM A 835 (2016), 186-225
                             WWW : http://geant4.org/
**************************************************************
```

nain4/n4_ui.cc
jacg: I had. Don't know where it went.

gonzaponte (approved):
We should start thinking about CLI tests, but that's for another PR.

jacg:
... and using a CL arg parsing library.

PR #57, opened 2023-07-31
Raise an error if 0 or too many world volumes are created

### LINKED ISSUE #54: Give an error if more than one volume is the world volume
It's easy to forget to write the `.in(...)` method when using `n4::place` to write a geometry. If this happens, no error is given, but the geometry is not fully constructed. With a complex geometry, this can lead to a headache. We should give an error when this happens to warn the user.

gonzaponte:
Closes #54 

Adds a check at run manager initialization time for world volume validity. There must be exactly 1 world volume (defined as a physical volume without a mother volume).

Tests fail intentionally for demonstration purposes.

### DIFF
--- a/nain4/n4_run_manager.cc
+++ b/nain4/n4_run_manager.cc
@@ -1,7 +1,39 @@
 #include "n4_run_manager.hh"
+#include "G4VPhysicalVolume.hh"
+#include "G4PhysicalVolumeStore.hh"
 
 namespace nain4 {
 
 run_manager* run_manager::rm_instance   = nullptr;
 bool         run_manager::create_called = false;
+
+
+void check_for_multiple_world_volumes() {
+  auto store = G4PhysicalVolumeStore::GetInstance();
+  std::vector<G4String> names{};
+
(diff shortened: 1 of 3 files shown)


jacg:
1. Have you got a plan for how to get Catch2 to assert the failure? Or are these human-inspection tests, like the overlap tests we did earlier?
2. There are tests for demonstrating that too few and too many world volumes give an error. It would be reassuring to see a test that demonstrates that the correct number of volumes gives no error.

gonzaponte:
>  1. Have you got a plan for how to get Catch2 to assert the failure? Or are these human-inspection tests, like the overlap tests we did earlier?

For now, they are based on human-inspection. We have a few instances of tests that check for things that use `exit(EXIT_FAILURE)` (like these ones) which is pretty difficult to test with Catch2. Perhaps it's time to think about exceptions.

> 2. There are tests for demonstrating that too few and too many world volumes give an error. It would be reassuring to see a test that demonstrates that the correct number of volumes gives no error.

I always forget to test the successful case. I will add it.

PR #62, opened 2023-08-04
More methods and tests for material properties

gonzaponte:
I've added more methods for material properties. Namely, those that:
- Take a single key to copy from another table: This applies to both methods: `add` and `NEW`
- Take an initializer list (necessary to resolve the ambiguity between the constructors of `std::vector<std::string>` and `std::string`).

But the main reason for this PR is to add tests to `n4::material_properties`. I've added a single test with many sections. Please let me know what you think.


jacg:
Excellent.

I wonder to what extent we can use some of these tests in the reference documentation. In some cases they are exhausting to read, most most of these are pretty digestible, I'd say.

PR #67, opened 2023-08-05
Find tag automatically in client-side test

gonzaponte:
This avoids the recurring `Bump tag to vX.Y.Z` commit

### DIFF
--- a/client_side_tests/client_fetch_content/CMakeLists.txt
+++ b/client_side_tests/client_fetch_content/CMakeLists.txt
@@ -2,10 +2,7 @@ cmake_minimum_required(VERSION 3.25)
 
 project(Client)
 
-execute_process(
-  COMMAND bash -c "git config --get remote.origin.url | cut -d '/' -f 4 | tr -d '\n'"
-  OUTPUT_VARIABLE USERNAME
-)
+set(USERNAME $ENV{GITHUB_USERNAME})
 
 execute_process(
   COMMAND bash -c "git log -1 | head -1 | cut -d ' ' -f 2 | tr -d '\n'"
--- a/client_side_tests/client_fetch_content/CMakeLists.txt
+++ b/client_side_tests/client_fetch_content/CMakeLists.txt
@@ -3,11 +3,17 @@ cmake_minimum_required(VERSION 3.25)
 project(Client)
 
 execute_process(
-  COMMAND bash -c "git tag | tail -1 | tr -d '\n'"
-  OUTPUT_VARIABLE TAG
+  COMMAND bash -c "git config --get remote.origin.url | cut -d '/' -f 4 | tr -d '\n'"
+  OUTPUT_VARIABLE USERNAME
 )
 
-message("Using Nain4 tag: ${TAG}")
+execute_process(
+  COMMAND bash -c "git log -1 | head -1 | cut -d ' ' -f 2 | tr -d '\n'"
+  OUTPUT_VARIABLE COMMIT_HASH
+)
+
+message("Using username: <${USERNAME}>")
+message("Using commit: <${COMMIT_HASH}>")
 
 # ANCHOR: fetch
 include(FetchContent)
@@ -16,8 +22,8 @@ set(FETCHCONTENT_UPDATES_DISCONNECTED ON CACHE BOOL "Cache package contents to a
 
 FetchContent_Declare(
   Nain4
-  GIT_REPOSITORY https://github.com/jacg/nain4.git
-  GIT_TAG        ${TAG}
+  GIT_REPOSITORY https://github.com/${USERNAME}/nain4.git
+  GIT_TAG        ${COMMIT_HASH}
   # make sure that no other nain4 installation is used
   OVERRIDE_FIND_PACKAGE
   SOURCE_SUBDIR nain4
--- a/client_side_tests/client_fetch_content/CMakeLists.txt
+++ b/client_side_tests/client_fetch_content/CMakeLists.txt
@@ -2,6 +2,13 @@ cmake_minimum_required(VERSION 3.25)
 
 project(Client)
 
+execute_process(
+  COMMAND bash -c "git tag | tail -1 | tr -d '\n'"
+  OUTPUT_VARIABLE TAG
+)
+
+message("Using Nain4 tag: ${TAG}")
+
 # ANCHOR: fetch
 include(FetchContent)
 
@@ -10,7 +17,7 @@ set(FETCHCONTENT_UPDATES_DISCONNECTED ON CACHE BOOL "Cache package contents to a
 FetchContent_Declare(
   Nain4
   GIT_REPOSITORY https://github.com/jacg/nain4.git
-  GIT_TAG        v0.1.9
+  GIT_TAG        ${TAG}
   # make sure that no other nain4 installation is used
   OVERRIDE_FIND_PACKAGE
   SOURCE_SUBDIR nain4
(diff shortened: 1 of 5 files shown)


client_side_tests/client_fetch_content/CMakeLists.txt
```diff
@@ -2,6 +2,13 @@ cmake_minimum_required(VERSION 3.25)
 
 project(Client)
 
+execute_process(
+  COMMAND bash -c "git tag | tail -1 | tr -d '\n'"
```
jacg, line 6: This assumes that `bash`, `git`, `tail` and `tr` are available.

Given that we only really run this in CI tests, and never on a real client's behalf, I guess we can ensure that this is true, for all practical purposes.

gonzaponte: For some reason that line returns nothing on GHA. That's why the tests are failing.

gonzaponte: `git tag` and its variants return nothing on GHA. I've tried several things but nothing seems to work. Do you have any ideas that I can try?

jacg: I guess it's because the repo cloned on GHA doesn't contain any tags ... until you explicitly fetch them.

gonzaponte: Thanks for the tip and apologies for the spam. This is finally ready.

jacg: How can you be sure that the tag you want will always be the last one?

+ Does git provide any guarantees about the order?
+ What happens if we add some other kinds of tags?

You should probably grep for `v.*.*` (sorry, that's glob syntax: I don't speak regex fluently (it's too much like line noise, for my liking)) and sort. (We can guarantee the presence of `grep` or `rg` with in the flake.)

client_side_tests/client_fetch_content/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.25)
 project(Client)
 
 execute_process(
-  COMMAND bash -c "git tag | tail -1 | tr -d '\n'"
+  COMMAND bash -c "git tag --sort=version:refname | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+' | tail -1 | tr -d '\n'"
```
jacg, line 6: Can have multiple digits between the dots. For example `v1.23.0` ... and I fully expect to get to `v0.1.10` pretty soon.

Also, `v.1.23.0-beta-6` is reasonable. But do we want to update to that automatically?

Also what happens if you are pushing to GHA while developing, before having merged and created the tag? I guess you have to remember to change `CMakeLists.txt` to point at your feature branch, and, here's the difficult bit **REMEMBER TO CHANGE IT BACK WHEN FINISHED**. Hmm, maybe we can make some Git hook deal with it.

gonzaponte: > Can have multiple digits between the dots

The `+`   in `[0-9]+` takes care of that. I tested it locally. REs are awful.

> But do we want to update to that automatically?

I don't know. At the moment our approach towards tagging has been pretty basic. We might decide to change it later.

> maybe we can make some Git hook deal with it.

That would be the best approach.

jacg: > The `+` in `[0-9]+` takes care of that.

Sorry for the noise: I knew that. Must have been asleep, or ...

>  REs are awful.

... a primordial aversion to RE made me not pay attention.

> > maybe we can make some Git hook deal with it.
> 
> That would be the best approach.

IIUC, as it stands, we have a problem: the client side tests will not be using the correct tag on GHA an feature branches. So this should not be merged. Or am I missing something?

gonzaponte: It depends on what we want to achieve. So far this test was only ensuring that the latest tag was working, it didn't say absolutely anything about the current branch. This PR only automates the tag selection.

If instead of doing that we want to test the current branch, then this needs to change our approach and this PR is not doing anything relevant.

jacg: > this test was only ensuring that the latest tag was working

The point of having CI tests run on feature branches is to warn you that the feature branch breaks something, before you merge it.

Anything that has already been merged should (in principle) pass CI tests every time, forever. So it's the CI failures on *feature* branches which give you interesting information. So a test that only verifies that something that has not only been merged but also tagged is still OK, seems rather pointless.

I understand the point of this PR to be absolving us of the need to change that tag by hand.

Given that that tag only appears in a client-side-test, I don't think that there should be a tag there at all! It should just use the branch that is being tested.

If we distribute a tag in some template for use by a human client (as opposed to CI) *that* tag *should* be updated to the latest stable one, preferably automatically.

gonzaponte: > If we distribute a tag in some template for use by a human client (as opposed to CI) that tag should be updated to the latest stable one, preferably automatically.

This made me remember why it was done like this. It was part of the documentation as the recommended way to incorporate nain4 in a project. Thus, we specified a tag. This means that the documentation cannot simply take the snippet from the test. It must either handle that particular line separately or pay the price of having a duplication.

For the tests, I agree, they must use the feature branch.

jacg: > It was part of the documentation

Aaaaaaah yes!

And yet we want the documentation tested. Bugger. Duplication, I guess. But then we'll get into a Catch-22 when the feature branch breaks the client-side test, which, by definition happens before we have tagged it.

gonzaponte:
I'm stuck. The CI creates some sort of merge commit. When I try to use the commit in fetch content it doesn't find it even though I can see it when I go to `https://github.com/jacg/nain4/commit/<commit_hash>`.

jacg:
Why don't you stop banging your head against this one and do something less soul-crushing for a while?

gonzaponte:
I am stubborn, but I am also doing other things in parallel while I figure this one out.

jacg:
> I'm stuck. The CI creates some sort of merge commit. When I try to use the commit in fetch content it doesn't find it even though I can see it when I go to `https://github.com/jacg/nain4/commit/<commit_hash>`.

Are you running into some problem related to CI almost certainly cloning with `--depth 1`?

gonzaponte:
Yes. But I'm not sure yet that cloning with `--depth 2` will solve the problem. If it did, is this something you would have to change in the CI configuration?

jacg:
I guess it would have to change in the checkout action ([`actions/checkout@v3.5.3`](https://github.com/actions/checkout)).

But even if it clones with `--depth 1`, then that 1 should be your latest commit ... unless they're doing something really weird.

gonzaponte:
> unless they're doing something really weird.

They are. The latest commit within the CI is this: b7afd910f4190568f7386a05a0a0f35b084fa7fd

gonzaponte:
Can you see the logs in this PR? Even since the GHA issue last week I've had plenty of problems fetching the logs.

jacg:
You mean the GHA logs? I can see these: https://github.com/jacg/nain4/actions/runs/5895187489

You'll have to refresh my memory about which GHA issue last week.

gonzaponte:
> I can see these

I should be more specific. Can you *load* the logs and see (for example) where it failed? I can't load the logs.

> You'll have to refresh my memory about which GHA issue last week.

It happened while you were offline. There was an issue with GHA and t would run the jobs, but I couldn't load the logs.

gonzaponte:
> I can't load the logs

It gets stuck like this
imaxe

jacg:
If I click on one of the `Details` links at the bottom of this PR, it takes me straight to the place in the log where it fails.

gonzaponte:
Ok, so it's on my side... somehow...

gonzaponte:
I got it working in GHA. We will need to discuss what's the best way to run the client-side tests offline.

PR #69, opened 2023-08-05
Break up monolithic flake

jacg:
The current flake that comes with `nain4` conflates (for historical reasons) provision of a number of different things, and there are a few things that it should provide but does not.

Can these be mapped to the [flakes output schema](https://nixos.wiki/wiki/Flakes#Output_schema)? :

1. `packages .<system>.nain4` and `.default`: `nain4` itself
2. `devShells.<system>.nain4` and `.default`: a development environment for working on `nain4`
3. `checks   .<system>.nain4`: everything necessary for testing `nain4` on CI. Could it just be executed in the workflow?
4. `devShells.<system>.client`: (basis of) a development environment for projects that use `nain4` to write applications in `Geant4`
5. `???` everything needed by users of projects written with `nain4`+`Geant4`, i.e. something that projects in category 4 would provide to their users.
6. `templates.default`: a template for starting projects in category 4. Could provide a number of these, with different pre-configured physics lists, sample detector/generators etc.
7. `packages.<system>.geant4`: `Geant4` with GUI, without threading. Could provide a number of these with different configurations.

I imagine the following differences between the above

+ 3 vs 2
  - 2 should use the in-tree `nain4` sources directly; should 3 get `nain4` via 1?
  - 2 is probably 3 plus LSP servers, debuggers, profilers etc.

+ 4 vs 2
  - 2 should use the in-tree `nain4` sources directly; 4 should get `nain4` via 1
  - 4 may need extra dependencies which are irrelevant to 2

+ 5
  - may need extra dependencies which are irrelevant to the others
  - can do without the testing framework which is needed by 2,3 and 4
  - probably doesn't belong in this flake, because it needs to be completed by the authors of that project

All this will need to work on Linux, MacOS with Apple Silicon and WSL2, but that’s a second-order issue at this point.


jacg:
This is essentially done, with the `bootstrap-client-project` app in the `nain4` flake.

We still have to look into putting `checks` to good use, which has been extracted into its own issue: #126.

gonzaponte:
Perhaps as the project grows we should split it even further. For instance with apps being defined in their own files. But we'll cross that bridge when we get there.

Issue #17, opened 2023-07-12
Consider running GHA with gcc in addition to clang


jacg:
Does anyone use `gcc` on MacOS?

jacg:
We got an error when we bumped the version of `nain4` in `GaP` and the user was using `gcc`. `clang` was perfectly happy.

IIRC, it was two `using CLHEP::pi`s in the same scope.

gonzaponte:
Yes, fixed [here](https://github.com/jacg/nain4/commit/dc6fc98776ccbd26555a743cd7b3000129d34d86)

jacg:
The point being that we have a concrete example of why we need to run `gcc` on GHA.

PR #82, opened 2023-08-16
Habemus Macam

### LINKED ISSUE #15: Sort out MacOS problems
Bizarre MacOS failures in `cmake` and in GHA, are diverting the time and energy of the only 2 people currently contributing to the project. As neither of these people have access to a machine running MacOS, it's difficult to make any progress on trying to fix these problems.

Yes, getting `nain4` to work smoothly for people using MacOS is *crucial*, but I don't think it's a good use of the 2 developers' time, at this very moment, to divert our limited resources away from all the other work that needs to be done urgently.

I will therefore allow the GHA MacOS job to fail, and I will ignore all MacOS problems until some progress has been made in other areas, especially getting the most important parts of the documentation written and published.

If anyone who actually *has* a Mac wants to have a go at working on this, that would be very welcome, and we could offer help. But without access to a Mac, this simply has to be given a much lower priority, for a while.

jacg:
Closes #15.

The remaining problems against which we were banging our heads recently, were caused by linking multiple versions of libc++, which was caused by the dependencies (G4 and Qt) and the client code using different compiler configurations.

The solution[^1] is to use the gcc environment rather than the clang one in the flake's `devShell`. This is only necessary on Darwin, but (for now) it has been changed on Linux too.

The change to gcc breaks the compile-time tests we added recently (which have been removed, to make GHA pass).

Thus, before we can merge this, we have to:

+ [x] Fix failing compile-time tests
+ [x] Revert to clang on Linux

@soleti this has been tested on macOS on Intel, but we do not have access to Apple Silicon, so it would be good if you could check that it works there too.

(Why is the macOS GHA so slow? The one I've been waiting for before submitting this PR, completed on Linux in under 9 minutes, the macOS one chugged along for muuuch longer and then failed spuriously, so I had to restart it. Then it took another 21 minutes.)

[^1]: This solution is temporary: a fix is [in the pipeline](https://github.com/NixOS/nixpkgs/issues/234710) upstream.

### DIFF
--- a/client_side_tests/client_fetch_content/CMakeLists.txt
+++ b/client_side_tests/client_fetch_content/CMakeLists.txt
@@ -10,7 +10,7 @@ set(FETCHCONTENT_UPDATES_DISCONNECTED ON CACHE BOOL "Cache package contents to a
 FetchContent_Declare(
   Nain4
   GIT_REPOSITORY https://github.com/jacg/nain4.git
-  GIT_TAG        v0.1.8
+  GIT_TAG        origin/fix-darwin
   # make sure that no other nain4 installation is used
   OVERRIDE_FIND_PACKAGE
   SOURCE_SUBDIR nain4
--- a/client_side_tests/client_fetch_content/CMakeLists.txt
+++ b/client_side_tests/client_fetch_content/CMakeLists.txt
@@ -10,7 +10,7 @@ set(FETCHCONTENT_UPDATES_DISCONNECTED ON CACHE BOOL "Cache package contents to a
 FetchContent_Declare(
   Nain4
   GIT_REPOSITORY https://github.com/jacg/nain4.git
-  GIT_TAG        origin/fix-darwin
+  GIT_TAG        v0.1.8
   # make sure that no other nain4 installation is used
   OVERRIDE_FIND_PACKAGE
   SOURCE_SUBDIR nain4
--- a/flake.nix
+++ b/flake.nix
@@ -12,7 +12,6 @@
     flake-utils.lib.eachSystem ["x86_64-linux" "i686-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin"] (system:
       let
         pkgs = import nixpkgs { inherit system; };
-
         my-geant4 = (pkgs.geant4.override {
           enableMultiThreading = false;
           enableInventor       = false;
@@ -45,14 +44,9 @@
           clang-tools
         ];
 
-        my-mkShell = pkgs.mkShell.override {
-          stdenv = if pkgs.stdenv.isDarwin then pkgs.                stdenv
-                                           else pkgs.llvmPackages_16.stdenv;
-        };
-
       in {
 
-        devShell = my-mkShell {
+        devShell = pkgs.mkShell {
           name = "G4-examples-devenv";
 
           packages = my-packages;
--- a/flake.nix
+++ b/flake.nix
@@ -36,8 +36,8 @@
             geant4.data.G4SAIDDATA
             geant4.data.G4PARTICLEXS
             geant4.data.G4NDL
-            clang_16
-            clang-tools
+            # clang_16
+            # clang-tools
             cmake
             cmake-language-server
             catch2_3
--- a/flake.nix
+++ b/flake.nix
@@ -22,29 +22,34 @@
           enableRaytracerX11   = false;
         });
 
+        my-packages = with pkgs; [
+          my-geant4
+          geant4.data.G4PhotonEvaporation
+          geant4.data.G4EMLOW
+          geant4.data.G4RadioactiveDecay
+          geant4.data.G4ENSDFSTATE
+          geant4.data.G4SAIDDATA
+          geant4.data.G4PARTICLEXS
+          geant4.data.G4NDL
+          cmake
+          cmake-language-server
+          catch2_3
+          just
+          gnused # For hacking CMAKE_EXPORT stuff into CMakeLists.txt
+          mdbook
+        ] ++ lib.optionals stdenv.isDarwin [
+
+        ] ++ lib.optionals stdenv.isLinux [
+          clang_16
+          clang-tools
+        ];
+
       in {
 
         devShell = pkgs.mkShell {
           name = "G4-examples-devenv";
 
-          packages = with pkgs; [
-            my-geant4
-            geant4.data.G4PhotonEvaporation
-            geant4.data.G4EMLOW
-            geant4.data.G4RadioactiveDecay
-            geant4.data.G4ENSDFSTATE
-            geant4.data.G4SAIDDATA
-            geant4.data.G4PARTICLEXS
-            geant4.data.G4NDL
-            # clang_16
-            # clang-tools
-            cmake
-            cmake-language-server
-            catch2_3
-            just
-            gnused # For hacking CMAKE_EXPORT stuff into CMakeLists.txt
-            mdbook
-          ];
+          packages = my-packages;
 
           G4_DIR = "${pkgs.geant4}";
           G4_EXAMPLES_DIR = "${pkgs.geant4}/share/Geant4-11.0.4/examples/";
--- a/flake.nix
+++ b/flake.nix
@@ -24,7 +24,7 @@
 
       in {
 
-        devShell = pkgs.mkShell.override { stdenv = pkgs.clang_16.stdenv; } {
+        devShell = pkgs.mkShell {
           name = "G4-examples-devenv";
 
           packages = with pkgs; [
--- a/flake.nix
+++ b/flake.nix
@@ -22,6 +22,23 @@
           enableRaytracerX11   = false;
         });
 
+        # Should be able to remove this, once https://github.com/NixOS/nixpkgs/issues/234710 is merged
+        clang_16 = if pkgs.stdenv.isDarwin
+          then pkgs.llvmPackages_16.clang.override rec {
+            libc = pkgs.darwin.Libsystem;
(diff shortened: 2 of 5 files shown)


soleti:
This looks like it's working on macOS Silicon as well, I am able to run `g4-examples/B1`.

jacg:
@soleti Could you also try the version in the `fix-clang-on-darwin` branch?

soleti:
Also works.

jacg:
This branch uses clang on Linux, and stdenv (g++) on Darwin. Over in `fix-clang-on-darwin` we have a version which uses clang on both, but forces the Darwin one to use the same version of libc++ as everything else. The latter is a bit ugly, and should become obsolete once the aforementioned [upstream fix](https://github.com/NixOS/nixpkgs/issues/234710) lands.

Which one do we prefer?

(I'll add the `fix-clang-on-darwin` solution here, for archival purposes: we can revert it if we want to stick with what is already here.)

`fix-clang-on-darwin` should sidestep the compile-time test failures on g++, but maybe we want to take this opportunity to address those. But I wouldn't delay merging this for the sake of fixing compile-time tests on g++.

g++ also spews an absurd amount of warnings.

There is also #17.

jacg:
> `fix-clang-on-darwin` should sidestep the compile-time test failures on g++

[Apparently](https://github.com/jacg/nain4/actions/runs/5888404966/job/15969543216), not.

gonzaponte:
> Which one do we prefer?

Will it still be ugly once the upstream fix is in place?

There is one advantage of keeping g++ on darwin and that's that we don't ignore g++, which we have been doing for a while. As you say:

> There is also https://github.com/jacg/nain4/issues/17.

> But I wouldn't delay merging this for the sake of fixing compile-time tests on g++.

When I get some time I will explore what's going on with that in a separate branch

> g++ also spews an absurd amount of warnings.

I guess this has to do with the different interpretations of `-Wshadow` and so on? If so we will fix it with pragmas like we did before.

jacg:
> Will it still be ugly once the upstream fix is in place?

Once the upstream fix is in place, it should look exactly like it did before we touched anything, but now it should work.

Currently the clang 16 env on Darwin uses a different version of libc++ from everything else, that's what was causing our problems. After the fix, it should use the same version. IIUC. So our old flake should Just Work.

> There is one advantage of keeping g++ on darwin and that's that we don't ignore g++, which we have been doing for a while. As you say:
> 
> > There is also #17.

I guess we should just add it to GHA.

Off the top of my head, it should be possible to have a default devShell with clang, and a second one with gcc. Then it should be possible to invoke the latter explicitly with something like `nix develop .#gcc`.

jacg:
> > `fix-clang-on-darwin` should sidestep the compile-time test failures on g++
> 
> [Apparently](https://github.com/jacg/nain4/actions/runs/5888404966/job/15969543216), not.

Fixed.

I've left the gcc-on-darwin-clang-on-linux solution in the history, for reference, and replaced it with the patch-clang-on-darwin-to-use-libc++-11 approach.

This avoids using g++, so the compile-time tests now pass.

I suggest that we merge, and deal with g++ related issues (broken compile-time tests, running g++ on GHA) in a separate PR.

client_side_tests/client_fetch_content/CMakeLists.txt
```diff
@@ -24,6 +24,14 @@ include(${Geant4_USE_FILE})
 
 add_executable(client_exe main.cc)
 
+if(APPLE)
+    target_link_options(Nain4 PRIVATE -undefined dynamic_lookup)
+endif()
+
+if(APPLE)
+    target_link_options(client_exe PRIVATE -undefined dynamic_lookup)
+endif()
```
gonzaponte, line 33: why is Nain4 a target here, this is a client, shouldn't need to mess with Nain4's link options.

client_side_tests/client_subdirectory/CMakeLists.txt
```diff
@@ -11,6 +11,14 @@ include(${Geant4_USE_FILE})
...
 add_executable(client_exe main.cc)
 
+if(APPLE)
+    target_link_options(Nain4 PRIVATE -undefined dynamic_lookup)
+endif()
+
+if(APPLE)
+    target_link_options(client_exe PRIVATE -undefined dynamic_lookup)
+endif()
+
```
gonzaponte, line 21: same remark as above.

flake.nix
```diff
@@ -22,29 +22,51 @@
...
+          cmake
+          cmake-language-server
+          catch2_3
+          just
+          gnused # For hacking CMAKE_EXPORT stuff into CMakeLists.txt
+          mdbook
+        ] ++ lib.optionals stdenv.isDarwin [
+
+        ] ++ lib.optionals stdenv.isLinux [
+        ];
```
gonzaponte, line 62: This is not necessary anymore, right?

jacg: Not necessary, but I want to keep it there, to show how it can be done.

client_side_tests/client_fetch_content/CMakeLists.txt
jacg: I don't know *why* it's needed, but [here is the proof](https://github.com/jacg/nain4/actions/runs/5890431218/job/15975511647) *that* it is needed.

gonzaponte: Ah, I think we go back again to the issue where fetchcontent grabs the last tag, not the current branch.
#67

client_side_tests/client_subdirectory/CMakeLists.txt
jacg: This one does seem to be unnecessary.

client_side_tests/client_fetch_content/CMakeLists.txt
jacg: Ugh.

gonzaponte: Can we have a commit changing `GIT_TAG` to your branch to convince ourselves that it's the only issue?

gonzaponte: It does. It confirms what I was thinking. Now it fails in the recompilation stage because `GIT_TAG` is not a tag and therefore it attempts to use the network. I'm confident that the tests pass without the `Nain4` link option (as they should). We need to address #67 to avoid future problems like this.

gonzaponte: So we revert back the `GIT_TAG` change and this will be ready to merge, I believe.

jacg: So we tag and merge?

gonzaponte: Ah wait, we need to update the documentation, right?

jacg: Uhhh ... which part? I've got so many context switches going on right know that I have no idea whatsoever what I am doing any more.

gonzaponte: https://jacg.github.io/nain4/how-to/enable-nain4-in-cmake.html

Although technically, dealing with stupid macos is not a nain4 problem. But I guess it's worth the trouble.

jacg: So we'd have to add the if-apple-then-dynamic-lookup gubbins somewhere in there? 

Yes, that absolutely needs to go in.

Anything else?

jacg: > dealing with stupid macos is not a nain4 problem

Unfortunately, in the harsh Real World, it is very much our problem.

gonzaponte: I can't think of anything else. Just remember to revert [9d9d552](https://github.com/jacg/nain4/pull/82/commits/9d9d5526f00d216b1ee51cf6a2b92f907ced797d).

We can provide a CMake function for our clients to do if-apple-then-dynamic-lookup more conveniently, but that will probably not work out of the box and deserves its own PR.

gonzaponte: The macos build is passing the tests. This means that the networkless recompilation test is not working as expected. My guess is that the command we use to restrict network usage is not working on macos. This means that it's allowed to use the network, fetch the branch and recompile, thus passing the test. But this is not supposed to happen.

gonzaponte: And, [I can confirm that this is the issue](https://github.com/jacg/nain4/actions/runs/5890975701/job/15977141731?pr=82#step:8:121):

> sudo: iptables: command not found

jacg:
Do you want to review #86 and/or #84 before I merge them?

I'm bored of having to cancel a duplicate workflow with every push, so it would be good to get at least #86 out of the way quickly.

gonzaponte:
There seems to be a linking problem in macOS now?

gonzaponte:
> But it is needed

I see two options:
+ Now that we understand the failures, we remove the offending lines, merge this PR and tag the new master
+ We merge this PR with the offending lines, tag master at that point and push another commit removing the lines.

I don't have a strong preference.

jacg:
What do you consider to be the offending lines? I don't think there are any (unless you're referring to the apple dynamic links in the client-side tests, whose removal I have finally managed to push, so they are not there any more).

If I've got that right, then we

1. change the tag in the client-side test
2. merge
3. tag

done.

gonzaponte (approved):

jacg:
You or me? I can do it, if you haven't started yet.

gonzaponte:
Go for it.

PR #84, opened 2023-08-17
Gcc devshell

### LINKED ISSUE #17: Consider running GHA with gcc in addition to clang

jacg:
Adds a second devShell which uses gcc instead of clang. This means we can now do

1. `nix develop <path-to-flake>` 
2. `nix develop <path-to-flake>#gcc`
3. `nix develop <path-to-flake>#clang`

1&3 give a clang environment, 2 gives a gcc environment.

This could be useful for users who have a strong preference, but, above all, for allowing us to test both gcc and clang on GHA.

However, I'm unsure about how we should organize the workflows? Specifically, should the gcc and clang tests be run in

1.  the same job
2. two separate jobs

?

+ Pro 1
   - trivial to set up
   - uses less CPU and network (doesn't need to boostrap the GHA environment twice)
+ Pro 2
   - reports gcc failures separately from clang failures
   - runs in parallel, so completes more quickly

Note: this sits on top of the as-yet-unmerged #82.

Closes #17.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,14 +19,15 @@ jobs:
         )
 
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.allow-fail }}
+    #continue-on-error: ${{ matrix.allow-fail }}
+    continue-on-error: ${{ matrix.devshell == 'gcc' }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04, macos-12]
         py: [311]
         devshell: [ clang, gcc ]
-        allow-fail: [false]
+        #allow-fail: [false]
 
     steps:
       - uses: actions/checkout@v3.5.3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,19 +47,19 @@ jobs:
         run: nix develop
 
       - name: Test nain4
-        run: nix develop ${{ matrix.devshell }} -c just test-nain4
+        run: nix develop .#${{ matrix.devshell }} -c just test-nain4
 
       - name: nix-shell
         run: nix-shell --run just
 
       - name: Run client-side tests
-        run: nix develop ${{ matrix.devshell }}  -c just test-client-side
+        run: nix develop .#${{ matrix.devshell }}  -c just test-client-side
 
       - name: Run compile-time tests
-        run: nix develop ${{ matrix.devshell }}  -c just test-compile-time
+        run: nix develop .#${{ matrix.devshell }}  -c just test-compile-time
 
       - name: Run nain4 examples
-        run: nix develop ${{ matrix.devshell }}  -c just test-examples
+        run: nix develop .#${{ matrix.devshell }}  -c just test-examples
 
       - name: Run G4 examples
-        run: nix develop ${{ matrix.devshell }}  -c just g4-examples/run-examples-that-work
+        run: nix develop .#${{ matrix.devshell }}  -c just g4-examples/run-examples-that-work
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04, macos-12]
         py: [311]
+        devshell: [ clang, gcc ]
         allow-fail: [false]
 
     steps:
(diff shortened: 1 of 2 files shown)


jacg:
FFS ... `allow-fail` has meaning in the GHA matrix. Why am I miserably failing to find *any* documentation on it?

jacg:
... because it doesn't have meaning in GHA, I invented it as a convenient name (for use in `continue-on-error`) it would seem.

jacg:
We now have this matrix:
```yml
    continue-on-error: ${{ matrix.devshell == 'gcc' }}
    strategy:
      fail-fast: false
      matrix:
        os: [ubuntu-22.04, macos-12]
        py: [311]
        devshell: [ clang, gcc ]
```
so 4 jobs run (2 OSes x 2 compilers), and those on gcc are allowed to fail.

Curiously, the gcc job [only fails on Linux, not Darwin](https://github.com/jacg/nain4/actions/runs/5891503733/job/15978697184). Is this because on macOS we're compiling with clang, even when in the gcc environment?

One slight wart is that we run `nix-shell` on both compilers' jobs, but `nix-shell` always uses the default (clang). It should be easy to switch it off on gcc.

client_side_tests/client_fetch_content/CMakeLists.txt
```diff
@@ -24,6 +24,14 @@ include(${Geant4_USE_FILE})
 
 add_executable(client_exe main.cc)
 
+if(APPLE)
+    target_link_options(Nain4 PRIVATE -undefined dynamic_lookup)
+endif()
```
gonzaponte, line 29: Forgot to delete this one.

jacg: I've removed all commits that added/removed this to any of the client-side tests. But I guess that GHA is not going to tell us anything sensible about this.

jacg: But you've reverted the IP tables-equivalent commit. What happened there?

gonzaponte: > I've removed all commits that added/removed this to any of the client-side tests. But I guess that GHA is not going to tell us anything sensible about this.

I don't understand what you are saying here. If you have removed it, why is it still there?

> But you've reverted the IP tables-equivalent commit. What happened there?

The GHA runs were stalling, I think it was because of that.

jacg: > If you have removed it, why is it still there?

You can't see it, because I can't push, because you've pushed the reversal of the ip-tables thing since I rebased, and don't know that to do with that. That's why I need to know more about the ip-tables commit which was added and then reverted. Do we want it there, or not?

gonzaponte: I want to keep the addition/removal for myself so I know that I tried that, but that's already done. you can kill both commits. The problem is that one of the client-side tests in macOS passes because of the wrong reason. But we can maybe skip it for now?

jacg: It passes in spite of the `set -e` that you added?

gonzaponte: It seems so... I don't understand that... nor what was happening with the ip tables-alternative.

jacg: Pfrrrrrrrrrprrrrrrrrrfrrrrrrrpthrrr!

At this rate we will not write a single line of nain4 code this week. It's all been sysadmin, configuration and related meta-meta-tedium. Grrrrrr.

gonzaponte: I suggest we skip the test (it's not so crucial after all), open an issue and move on with our lives.

jacg: Just to be sure, we're talking about at test that is spuriously passing and has just been merged in #82?

Go ahead, open the issue, you're much clearer on this than I am.

gonzaponte:
> we're talking about at test that is spuriously passing and has just been merged in https://github.com/jacg/nain4/pull/82?

Yes.

jacg:
I've lost track. Do we want to do anything more with this, or shall we merge?

gonzaponte:
We want to skip the test, right?

jacg:
I must be missing something: I don't see what this test has to do with this PR. What's more, it seems to be *failing* on master.

gonzaponte:
>  I don't see what this test has to do with this PR

Ugh, nothing. I got lost in the many conversations. This can be merged.

> it seems to be failing on master.

Ugh x2. Now I'm puzzled.

jacg:
> I got lost in the many conversations.

Yup, me too.

gonzaponte:
> > it seems to be failing on master.
> Ugh x2. Now I'm puzzled.

I'm trying to undo the `set -e` in my remote. If it works I will push it to master and we will deal with that on #88. But that shouldn't stop this from being merged.

jacg:
I've just push a disabling of that test. It's getting in the way too much, right now.

gonzaponte:
> I'm trying to undo the set -e in my remote. If it works I will push it to master and we will deal with that on https://github.com/jacg/nain4/issues/88. But that shouldn't stop this from being merged.

I've also pushed this.

PR #86, opened 2023-08-17
Improve GHA workflow run conditions

### LINKED ISSUE #85: Stop GHA running twice when pushing to a PR
Currently, pushing to an open PR results in the same GHA actions being run twice. This is, quite simply, a pointless waste of energy that clogs up runner queues.

The solution should be pretty simple, there are a few alternative approaches, but I don't have the spare brain cycles to deal with this right now, so here's an issue lest we forget.

jacg:
Closes #85.


jacg:
It's not ideal: it leaves a cancelled workflow in the "All workflows" log, when two workflows are triggered by a push to a PR, but I couldn't find a quick workaround, and I don't think it's worth spending more time on it.

.github/workflows/test.yml
```diff
@@ -2,15 +2,22 @@ name: Build and Test
...
+jobs:
   build-and-test:
-    if: "! contains(github.event.head_commit.message, '[skip ci]')"
+
+    # See https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/
+    if: (! contains(github.event.head_commit.message, '[skip ci]')                   &&
+                  ((github.event_name                             != 'pull_request') ||
+                   (github.event.pull_request.head.repo.full_name !=
+                    github.event.pull_request.base.repo.full_name))
+        )
```
gonzaponte, line 19: Just to check that I understood the logic:
+ If you make a PR, two runs are triggered (push and PR), but it will skip the PR one because `event_name == pull_request` and `head == base`.
+ If anyone else makes a PR, the PR run will execute because `head != base`, but at the same time for the push run  `event_name != pull_request`, so it will execute too, will it not?

jacg: IIUC, the `*.repo_full_name` will depend on the repository in which the PR branch lives. This repo can't control (trigger/suppress) GHAs running in other repos, so the best it can do is to prevent duplication in this repo.

But this single workflow definition is seen by both repos, so maybe it could prevent duplicate runs across different repos. Given that we always want the receiving repo to get reports on the success of any commits to the PR, we want it to run there. But maybe we would like workflows not to be triggered in the contributing repo on any branch which is a PR in the receiving repo. Is it possible to detect this?

But maybe we don't want to suppress it at all, as then the workflows would appear to be (mysteriously) missing from the submitter's repo.

gonzaponte (approved):
You might want to remove the REVERTME commit and its counterpart.

jacg:
Not really. They kinda explain the need for some of the shenanigans. Or they did at some stage. Maybe not any more.

jacg:
You resolved the one open conversation, so shall I merge?

gonzaponte:
> You resolved the one open conversation

and approved the PR!

> so shall I merge?

Yes, please.

PR #91, opened 2023-08-17
Quick hack to fix compile-time tests on gcc

### LINKED ISSUE #89: Fix gcc compile-time tests on Linux
They are currently allowed to fail.

jacg:
Fixes #89.

Not sure it's worth while making it any better, at present.

### DIFF
--- a/compile-time-tests/generate_tests.py
+++ b/compile-time-tests/generate_tests.py
@@ -110,6 +110,16 @@ int main() {{
 }}
 """
 
+def found(needle, haystack):
+    return needle in sanitized(haystack)
+
+def sanitized(text):
+    return (text
+            .replace('’', "'")
+            .replace('‘', "'")
+            )
+
+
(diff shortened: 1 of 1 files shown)


compile-time-tests/generate_tests.py
```diff
@@ -110,6 +110,16 @@
 }}
 """
 
+def found(needle, haystack):
+    return needle in sanitized(haystack)
+
+def sanitized(text):
+    return (text
+            .replace('’', "'")
+            .replace('‘', "'")
```
gonzaponte, line 119: Replace replaces all occurrences by default.

jacg: Look carefully: they are different quotes.

gonzaponte: I swear I looked at them for a good minute and a half and didn't see the difference.

gonzaponte (approved):

PR #124, opened 2023-09-07
Automate nixGL usage

### LINKED ISSUE #60: Automate nixGL in justfile
This would require detecting if you are

1. running in nix
2. on non-NixOS (and non-MacOS, probably)
3. and trying to initialize the UI

in which case it should prepend some variation on the `nix --extra-experimental-features "nix-command flakes" run --impure github:guibou/nixGL --` theme, to the command being executed.

Not sure whether this would be more trouble than it's worth.

gonzaponte:
~Uses nixGL in Ubuntu and WSL, and ignores it in macOS and NixOS. The recipe prepends~
Uses nixGL if not running on macOS or NixOS. The recipe prepends

```bash
nix --extra-experimental-features "nix-command flakes" run --impure github:guibou/nixGL --
```
to `just run` when necessary.

syntax:

```bash
just view [ARGS] # executes `just run [ARGS]` with the appropriate GL configuration
```

Closes #60.


jacg:
Slight misnomer in your opening comment: this has nothing to do with Ubuntu. There is only

+ NixOS
+ macOS
+ everything else

More importantly, I don't think there should be a `view` recipe at all, only `run`. The whole point is that the user shouldn't have to think about which `just` recipe to use: the same recipe should work regardless of what CLI args are passed to the app.

However, we should anticipate that our need-for-nixGL detection might make a mistake in some cases, so we should provide alternative run-recipes which explicitly enable/disable nixGL:

+ `just run-with-nixgl`
+ `just run-without-nixgl`

I don't like those names. How about

+ `run-con-nixgl`
+ `run-sin-nixgl`

or

+ `run-avec-nixgl`
+ `run-sans-nixgl`

Also, I think that the `run` recipe should be the default in the client's app, but `test` should be default in `nain4` itself.

gonzaponte:
> I don't think there should be a view recipe at all, only run. The whole point is that the user shouldn't have to think about which just recipe to use:

Right. I was a bit misguided by the fact that now there are no positional arguments, but named arguments can still be passed to `just run` in the same way.

> I don't like those names

I don't either. How about a flag that forces nixgl?

`just run [--force-nixgl]` or simply `just run [--nixgl]` / `just run [nixgl=true]`

I haven't explored all the syntax available in `just` recipes, so I don't know if that's easy to implement.

> Also, I think that the run recipe should be the default in the client's app, but test should be default in nain4 itself.

Sounds reasonable.

jacg:
>> Also, I think that the run recipe should be the default in the client's app, but test should be default in nain4 itself.

> Sounds reasonable.

... except that, as we've seen in the #125, passing flags to a default recipe doesn't work, and the client's app needs at least one flag to do anything interesting.

jacg:
`view.sh` needs a different name. Something along the lines of 

`view.sh` should not be duplicated between `nain4` and the client app.

gonzaponte:
> Something along the lines of

Along the lines of...?

> is it even needed in nain4 itself?

Good point. Probably not, unless we introduce GUI-based tests.

PR #125, opened 2023-09-07
Prettify tests output

gonzaponte:
Implements a pytest-like test output with three different levels of verbosity:
+ `-vv`: full output, similar to what we have now:
```
Filters: "nain material"
Randomness seeded to: 2507992811
===============================================================================
All tests passed (47 assertions in 1 test case)

[...]
```

+ `-v`: just the name of each test
```
    nain material
    nain material_properties
    [...]
```

+ no flags: `.` for success, `F` for failure
```
..........F........F.................

```

The green/red coloring is preserved in all three cases, and they still show the (colored) summary message at the end:
```
Ran 37 tests

===========================================================================
OVERALL: ============================== PASS ==============================
===========================================================================
```

Furthermore, the output of failed tests is shown just before the summary, for instance:
```
####################################################
# nain map
####################################################

Filters: "nain map"
Randomness seeded to: 523894240

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
nain4-tests is a Catch2 v3.3.2 host application.
Run with -? for options

-------------------------------------------------------------------------------
nain map
-------------------------------------------------------------------------------
/home/gonzalo/sw/git/nain4/nain4/test/test-nain4.cc:1650
...............................................................................

/home/gonzalo/sw/git/nain4/nain4/test/test-nain4.cc:1660: FAILED:
  CHECK( false )

===============================================================================
test cases: 1 | 1 failed
assertions: 3 | 2 passed | 1 failed
```

Unfortunately, I haven't found a way of passing flags when using the implicit default recipe (e.g.  `just -vv`), so running `just` will run the default behavior which is no flags. To apply the flags we need to do `just test-nain4 [FLAGS]`, which is a bit inconvenient.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,7 +58,7 @@ jobs:
         run: nix develop
 
       - name: Test nain4
-        run: nix develop .#${{ matrix.devshell }} -c just test-nain4
+        run: nix develop .#${{ matrix.devshell }} -c just test-nain4 -v
 
       - name: Test bootstrapping of client project
         run: |
--- a/justfile
+++ b/justfile
@@ -1,12 +1,13 @@
 # -*-Makefile-*-
 
+_ *FLAGS_AND_ARGS:
+    just test-nain4 {{FLAGS_AND_ARGS}}
+
 test-nain4 *FLAGS:
     just nain4/test-all {{FLAGS}}
 
-
 test-examples:
     just examples/
-
 
 test-client-side:
     just client_side_tests/test-all
--- a/justfile
+++ b/justfile
@@ -1,7 +1,7 @@
 # -*-Makefile-*-
 
-test-nain4:
-    just nain4/test-all
+test-nain4 *FLAGS:
+    just nain4/test-all {{FLAGS}}
 
 
 test-examples:
--- a/nain4/test/run-each-test-in-separate-process.sh.in
+++ b/nain4/test/run-each-test-in-separate-process.sh.in
@@ -15,7 +15,9 @@ EXE=@CMAKE_INSTALL_PREFIX@/bin/@TEST_EXECUTABLE@
 
 while read -r testname
 do
-    if ! $EXE "$testname" ${FLAGS}; then
+    OUTPUT=$($EXE "$testname" ${FLAGS})
+    STATUS=$?
+    if ! $STATUS; then
         FAILED=$FAILED"$testname"\\n
         NFAILED=$((NFAILED+1))
     else
--- a/nain4/test/run-each-test-in-separate-process.sh.in
+++ b/nain4/test/run-each-test-in-separate-process.sh.in
@@ -1,23 +1,22 @@
 #!/usr/bin/env sh
 
-if $1 -eq "-v"; then
-    verbose=true
-    shift
-else
-    verbose=false
-fi
 
+case $1 in
+    -vv) PATTERN=""; verbose=2; shift; FLAGS=$@ ;;
+    -v ) PATTERN=""; verbose=1; shift; FLAGS=$@ ;;
+    *  ) PATTERN=$1; case $2 in
+                         -vv) verbose=2; shift; shift; FLAGS=$@ ;;
+                         -v ) verbose=1; shift; shift; FLAGS=$@ ;;
+                         *  ) verbose=0;               FLAGS=$@ ;;
+                     esac ;;
+esac
 
-PATTERN=$1
-if [ ! -z $2 ]; then
-    shift
-    shift
-fi
-FLAGS=$@
 
 NPASSED=0
 NFAILED=0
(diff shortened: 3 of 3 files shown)


jacg:
*Very* nice. Very useful.

+ The `-vv` version is monochrome; I found the original green/yellow/default a bit more informative.

+ Should capture stderr too.

+ This problem existed before, but it stands out more now: the `======` markers are in the middle of tests, it would be much better if they separated tests.

+ Passing flags to the implicit default recipe results in the flags being eaten up by `just` itself. So the default recipe needs to be invoked by name. So how about giving it an almost invisible name, like `_`? I've pushed a POC.

jacg:
Do we want to make it more robust against commutation of `-v`?

See
+ `just test-nain4    '*enumerate*' -v`
+ `just test-nain4 -v '*enumerate*'   `
and
+ `just    test-nain4 -v`
+ `just -v test-nain4   `

gonzaponte:
> The -vv version is monochrome; I found the original green/yellow/default a bit more informative.

Yes, that was the price to pay. I couldn't find a way to keep the original coloring. Perhaps you have some ideas here.

> Should capture stderr too.

I think it does, doesn't it? If you add a failure to a test it doesn't complain immediately but only at the end of the run. Perhaps Catch2 doesn't write to stderr at all as I was thinking. In any case, that's easy to do.

> This problem existed before, but it stands out more now: the ====== markers are in the middle of tests, it would be much better if they separated tests.

Yes, but doing this with sh is painful at best. I don't want to add another dependency, especially if we want to provide this with the client app, but implementing all this in another language would be 100x more pleasant.

> So how about giving it an almost invisible name, like _?

I would also be happy with `t` as in `test`.

> Do we want to make it more robust against commutation of -v?

I thought of this. Again, it's getting silly with sh. Because the logic is complicated and needs to be implemented explicitly. I thought that the most natural syntax (or perhaps just the easiest to implement) would be with -v at the end and didn't bother to do better than that.

jacg:
> implementing all this in another language would be 100x more pleasant

Maybe this is a job for [`nu`](https://www.nushell.sh/)?

jacg:
> Passing flags to the implicit default recipe results in the flags being eaten up by just itself. 

Of course, this is precisely why `--` exists, but it doesn't work in this case. Not sure why.

gonzaponte:
> Maybe this is a job for [nu](https://www.nushell.sh/)?

`just` [supports this](https://github.com/casey/just#shell), so it could be a good solution, I will have to read it more carefully.

> > Passing flags to the implicit default recipe results in the flags being eaten up by just itself.

>  Of course, this is precisely why -- exists, but it doesn't work in this case. Not sure why.

Yup, I had also tried that. Probably worth an issue upstream, but I won't have time in the immediate future to contribute , if they ask for it.

Given the apparent limitations I propose the following:
+ `just t [ARGS]` runs the test suite
+ `just r [ARGS]` runs the main program
+ `just` runs the default recipe (either `r` or `t`) with some predetermined `ARGS`

jacg:
> Probably worth an issue upstream

Been there, [done that](https://github.com/casey/just/issues/1668), got the T-shirt.

gonzaponte:
> Been there, [done that](https://github.com/casey/just/issues/1668), got the T-shirt.

That was quick. I kinda agree... *if and only if* you use short recipe names.... for long, explicit names I don't think that's adequate. Particularly given that autocompletion doesn't work smoothly (for me).

jacg:
`t` and `r` are too cryptic. If you have typed `just r` you've had to make a conscious choice about running vs testing, and following it with an extra `un` is essentially zero overhead.

> > Been there, [done that](https://github.com/casey/just/issues/1668), got the T-shirt.
> 
> That was quick. I kinda agree...  _if and only if_ you use short recipe names.... for long, explicit names I don't think that's adequate. 

Exactly, which is why I chose `_`.

But it's an ad-hoc hack and, well, `--` is also an ad-hoc hack, but it's a de-facto standard ad-hoc hack, and implementing it in `just` itself would make it work

+ uniformly
+ without the justfile author having to think or do anything about it.

> Particularly given that autocompletion doesn't work smoothly (for me).

Yes, I'm sure that we can improve on that, but it's an Nth order problem.

Anyway, the whole point of the default recipe is that you don't have to make a conscious choice. You should be able to go into a project and simply type `just` without thinking about it, and something sensible (and harmless) should happen. If what happens isn't what you wanted, then you have to start thinking, but what just happened should give you some basis (or reminder) for the upcoming thought process.

I'd say that `just` (with nothing else) should

+ in the client repo: run the GUI with zero (or a few) events,
+ in the `nain4` repo: run the tests with intermediate verbosity.

PR #153, opened 2023-10-10
Random direction

jacg:
TODO:

+ [ ] Add chapters to the mdbook
+ [ ] Add in-code documentation comments to show up in LSP

### DIFF
--- a/doc/src/how-to/generate_random_numbers.md
+++ b/doc/src/how-to/generate_random_numbers.md
@@ -1,9 +1,10 @@
 # How to generate random numbers
 
-`nain4` offers a range of utilities to generate random numbers. This
-includes scalars, tuples and 3-vectors. All these methods are
-gathered in the namespace `nain4::random` and exported to
-the namespaces `nain4` and `n4`.
+`nain4` offers a range of utilities to generate random numbers. This includes
+scalars, tuples and 3-vectors. All these methods are declared in the
+[header](../reference/headers.md) `<n4-random.hh>` which is transitively
+included by `<n4-utils.hh>`. They are gathered in the namespace `nain4::random`
+and exported to the namespaces `nain4` and `n4`.
 
 
 ## Scalars
--- a/doc/src/how-to/generate_random_numbers.md
+++ b/doc/src/how-to/generate_random_numbers.md
@@ -1,0 +1,97 @@
+# How to generate random numbers
+
+`nain4` offers a range of utilities to generate random numbers. This
+includes scalars, tuples and 3-vectors. All these methods are
+gathered in the namespace `nain4::random` and exported to
+the namespaces `nain4` and `n4`.
+
+
+## Scalars
+
+The methods described in this section generate a single number.
+
+### Continuous distributions
+
+- `uniform()`: generates a random floating point number homogeneously
+  distributed in the range [0, 1).
+- `uniform(low, high)`: generates a random floating point number
+  homogeneously distributed in the range [`low`, `high`).
+- `uniform_half_width(hw)`: generates a random floating point number
+  homogeneously distributed in the range [`-hw`, `hw`).
+- `uniform_width(w)`: generates a random floating point number
+  homogeneously distributed in the range [`-w/2`, `w/2`).
+
+### Discrete distributions
+
+- `biased_coin(p_true)`: generates a random boolean with probability
+  `p_true` of being `true`.
+- `fair_die(n_sides)`: generates a random unsigned integer in the
+  range [0, `n_sides - 1`] with equal probability.
+- `biased_choice(weights)`: provides a generator of unsigned integers
+  in the range [0, `weights.size() - 1`] with given `weights`. The
+  generator must be called to obtain a number.
+
+
+## Tuples
+
+- `random_in_disc(r)`: generates a pair of floating point numbers `{x, y}` that satisfy `x^2 + y^2 <= r^2`.
+
+## 3-vectors
+
+The methods described in this section generate `G4ThreeVector`s.
+
+- `random_in_sphere(r)`: generates a vector of floating point numbers `{x, y, z}` that satisfy `x^2 + y^2 + z^2 <= r^2`.
+
+### Directions
+
+`nain4` provides the `direction` class to help generate random
+directions. The default state of this tool is to generate vectors
+homogeneously in 4π, but it can be configured to restrict the
+generation to certain directions. The following methods describe how
+to restrict the directionality. Unless otherwise stated, θ is the
+polar angle measured with respect to the positive z-axis and φ is the
+azimuthal angle measured with respect to the x axis
+counterclockwise. Use `.get()` to generate a random number.
+
+- `min_cos_theta(cos_th)`/`max_cos_theta(cos_th)`: set the
+  bounds for the cosine of theta angle such that `min_cos_theta <=
+  cos_theta < max_cos_theta`.
+
+- `min_theta(th)`/`max_theta(th)`: set the bounds for the theta angle such
+  that `min_theta <= theta < max_theta`.
+
+- `min_phi(p)`/`max_phi(p)`: set the bounds for the phi angle such that
+  `min_phi <= phi < max_phi`.
+
+- `axis`: rotate the reference frame such that the z-axis is aligned
+  with `axis`. θ will be measured with respect to the rotated z axis
+  and φ with respect to the rotated x axis (which I don't know how
+  it's defined. either x or y is `axis.cross(zaxis)`).
+
+- `bidirectional()`: accept both the current selection and its
+  fully-mirrored counterpart.
+
+- `exclude()`: invert the selection. It generates directions that do
+  not satisfy any of the selected criteria. Bear in mind that this
+  method is more computationally expensive as it needs to generate
+  directions until the result not rejected. Very restricted
+  directionalities are discouraged.
+
+#### Examples
+
+- Isotropic emission: `n4::direction()`
+
+- Opening angle wrt the positive z axis:
+  `n4::direction().max_theta(theta)`
+
+- Opening angle wrt the negative x axis:
+  `n4::direction().max_theta(theta).axis({-1, 0, 0})`
+
+- Only one octant:
+  `n4::direction().min_cos_theta(0).min_phi(CLHEP::halfpi).max_phi(CLHEP::pi)`
+
+- Back-to-back beams wrt the positive z axis with some opening angle:
+  `n4::direction().max_theta(theta).bidirectional()`
+
+- Isotropic except for some small angle around the negative z-axis:
+  `n4::direction().min_theta(7*CLHEP::pi/8).exclude()`
--- a/doc/src/how-to/generate_random_numbers.md
+++ b/doc/src/how-to/generate_random_numbers.md
@@ -45,32 +45,49 @@ The methods described in this section generate `G4ThreeVector`s.
 
 ### Directions
 
-`nain4` provides the `direction` class to help generate random
-directions. The default state of this tool is to generate vectors
-homogeneously in 4π, but it can be configured to restrict the
-generation to certain directions. The following methods describe how
-to restrict the directionality. Unless otherwise stated, θ is the
-polar angle measured with respect to the positive z-axis and φ is the
-azimuthal angle measured with respect to the x axis
-counterclockwise. Use `.get()` to generate a random number.
+`nain4` provides the `direction` builder to help generate random directions. By
+default this tool generates unit vectors distributed isotropically in 4π, but it
+can be configured to restrict the generation to certain directions.
 
-- `min_cos_theta(cos_th)`/`max_cos_theta(cos_th)`: set the
-  bounds for the cosine of theta angle such that `min_cos_theta <=
-  cos_theta < max_cos_theta`.
+The general usage pattern is
 
-- `min_theta(th)`/`max_theta(th)`: set the bounds for the theta angle such
-  that `min_theta <= theta < max_theta`.
+```c++
+auto generator = n4::random::direction{}.<optional extra specifications>;
+auto one_random_unit_vector = generator.get();
+```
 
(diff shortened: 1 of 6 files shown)


jacg:
Is there some rationale behind using `halfpi` instead of `pi` all over the place. I find things like `cos_pi8  = std::cos(CLHEP::halfpi/4);` a little perverse.

gonzaponte:
I think I got used to thinking in terms of pi/2 as I wrote the tests. Probably because halfpi is half of the limit of theta, so it was helping me visualize it. But I agree it's perverse.

jacg:
It might make sense in some contexts, especially when you're thinking in quadrants or octants, but for assignment to that kind of name, it was a bit much.

Also, the choice of order z,x,y in the generator names is ... interesting.

gonzaponte:
> Also, the choice of order z,x,y in the generator names is ... interesting.

Theta goes before phi, therefore z goes before x and y. And obviously, x goes before y.
That's just how it is. By law. :)

I tried different orders, this is the one that convinced me the most, but I am not too attach to it if you prefer a different one.

jacg:
Hmm, I've mostly replaced all that stuff with Cartesian base-2 counting order, but it's all up in the air at the moment. I'll push a WIP in a minute.

jacg:
Would you be happy to replace the `dir_bias` in the bidirectional test, with checks on the means of the generated data?

gonzaponte:
Sure.

jacg:
OK, what does `axis` do?

gonzaponte:
change the reference system so you point in one direction and select the rest of the parameters accordingly. For instance, in the test, I chose the cartesian axes, so the generator shoots beams in those directions with some opening angle.

jacg:
I don't understand.

This:

```c++
n4::random::direction{}.min_theta(pi/6)
```
gives these stats:

```
               min            max         spread           mean
x:       -0.999988       0.999999        1.99999    -0.00260523
y:       -0.999999       0.999978        1.99998      0.0012637
z:       -0.999971       0.866021        1.86599     -0.0660735
```

which suggests that, `ϑ=0` is on the positive z-axis. So far, so good.

Now, if I add `.exclude()`:
```c++
n4::random::direction{}.min_theta(pi/6).exclude()
```
I would expect it to give me the complement of the previous directions, so something like
```
               min            max         spread           mean
x:       -0.5            0.5             1.0               0
y:       -0.5            0.5             1.0               0
z:        0.866021       1.0             0.134             0
```

but it gives

```
               min            max         spread           mean
x:       -0.999938       0.999983        1.99992    -0.00309197
y:        -0.99999       0.499365        1.49935      -0.454033
z:       -0.999975       0.999994        1.99997      0.0590881
```
I'm especially perplexed by the asymmetry in *Y*.

gonzaponte:
You've clearly catched a bug. I will try to look into it.

gonzaponte:
phi from 3vector came in [-pi, pi) while I was assuming [0, 2pi).

gonzaponte:
I've done a bit of refactoring and fixing some tests that didn't make sense. Only 1 of them fails because of directional bias, and it feels like it's just a random fluctuation.

jacg:
I have strengthened the theta test. It fails with a sign discrepancy. Is this a bug in the test or the implementation (... or both)?

gonzaponte:
> Is this a bug in the test or the implementation (... or both)?

The test. You were using values above `pi` for theta (pushed a fix). But this is a good reminder that we need to add range checks to setters.

jacg:
Ah, good, range checking was going to be my next question.

We also need docs.

doc/src/how-to/generate_random_numbers.md
```diff
@@ -0,0 +1,97 @@
+# How to generate random numbers
+
+`nain4` offers a range of utilities to generate random numbers. This
+includes scalars, tuples and 3-vectors. All these methods are
+gathered in the namespace `nain4::random` and exported to
+the namespaces `nain4` and `n4`.
```
jacg, line 6: Not sure what you mean by this.

They are all in `nain4::random` and `n4` is a synonym for `nain4`, hence they are also available in `n4::random`.

"and exported to the namespaces `nain4` and `n4`" suggests that you could say something like `n4::uniform`, which is not the case.

jacg:
Ugh, I think I accidentally sent a single comment. Don't do anything for now, I'll be reworking it.

jacg:
Do you remember any interesting tools for drawing diagrams in mdbook?

Adding some pictures to these docs would help immensely.

jacg:
Wouldn't it be better to replace `.axis` with `.rot_{x,y,z}`?

Is there a concise way of dealing with the double-sipm use case of restricting to rectangular regions? `{min,max}_cos_theta` does half the job, it also requires something convenient for ... erm ...`{min_max}_cos(phi + 90 degrees)`.

Also, we need some convenient visualizer, so that the client can easily inspect the region that has been defined.

doc/src/how-to/generate_random_numbers.md
gonzaponte: This is wrong. I was assuming that everything in `n4::random` was being exported to the main namespace, but I didn't check it when I wrote it.

gonzaponte:
> Do you remember any interesting tools for drawing diagrams in mdbook?

Hmm. I don't remember, but I believe we looked for such a tool sometime between [big bang, now). I can't do it now, but I will look for it.

> Adding some pictures to these docs would help immensely.

I agree.

> Wouldn't it be better to replace .axis with .rot_{x,y,z}?

It would complement it. You should be able to orient it in any direction. I don't have a strong argument for the name. Whatever feels more self-explanatory.

> Is there a concise way of dealing with the double-sipm use case of restricting to rectangular regions? 

This is "tricky" because limits on cartesian coordinates in the real world depend on the distance. And in the unit vector they need to be calculated based on angles, which would not be very efficient. Not disastrous, but might not be worth the effort.

I believe in the particular case of double_sipm we have

```c++
n4::random::direction{}.axis({1, 0, 0}).max_theta(something easy to calculate).bidirectional()
```

which might (not sure) be a bit less efficient, but does the job quite well. If it's worth it, we can try to implement it, but we enter the realm of verifying that the selected options make sense.

> Also, we need some convenient visualizer, so that the client can easily inspect the region that has been defined.

It should also be easy to visualize with the GUI. Not sure if we can do better than that without a lot of effort.

doc/src/how-to/generate_random_numbers.md
```diff
@@ -0,0 +1,97 @@
...
+  `n4::direction().max_theta(theta)`
+
+- Opening angle wrt the negative x axis:
+  `n4::direction().max_theta(theta).axis({-1, 0, 0})`
+
+- Only one octant:
+  `n4::direction().min_cos_theta(0).min_phi(CLHEP::halfpi).max_phi(CLHEP::pi)`
+
+- Back-to-back beams wrt the positive z axis with some opening angle:
+  `n4::direction().max_theta(theta).bidirectional()`
```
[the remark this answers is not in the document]
gonzaponte: Yup, I agree. I was trying to not repeat words. I'm struggling to find the proper words to describe the bidirectional feature unambiguously.

jacg:
> Is there a concise way of dealing with the double-sipm use case of restricting to rectangular regions? `{min,max}_cos_theta` does half the job, it also requires something convenient for ... erm ...`{min_max}_cos(phi + 90 degrees)`.

I think we want the ability to specify bands whose thickness is given by (half)-tan of angles. The question is, what is best compromise between parsimony, generality, usefullness and ease of understanding?

I'm thinking along the lines of bands relative to equators around the three different axes. Though we could only allow specifying bands around one axis, and get the others by rotation operations. Here is an obvious place where the aforementioned compromises come into play.

Let's start by imagining having methods for bands around each axis, with most general methods. Then we can start paring it down. Also, I'll start with horrendously explicit names.

+ `{x,y,z}_equator_band_{min,max}_tan(ratio)`
+ `{x,y,z}_equator_band_symmetric_tan(ratio)`
+ `{x,y,z}_equator_band_symmetric_half_tan(ratio)`

probably covers everything. Do these names make sense?

The first bunch allow off-equator bands. So maybe `equator` could/should be dropped.

jacg:
> > Wouldn't it be better to replace .axis with .rot_{x,y,z}?
> 
> It would complement it. You should be able to orient it in any direction. I don't have a strong argument for the name. Whatever feels more self-explanatory.

The rotations would allow you to orient it in any direction. The trouble with `axis` is that it's not clear how the directions orthogonal to z are re-oriented, unless I'm missing something.

jacg:
> > Also, we need some convenient visualizer, so that the client can easily inspect the region that has been defined.
> 
> It should also be easy to visualize with the GUI. Not sure if we can do better than that without a lot of effort.

The G4 GUI is probably the best we can do in the sort term. We'll need some pre-built app that contains nothing but an `n4::random::direction` sample and pops up its result in the GUI.

Once we're a huge project with endless resources, we can create one of those web apps containing a code editor and an updated-as-you-type image next to it.

jacg:
> > Is there a concise way of dealing with the double-sipm use case of restricting to rectangular regions?
> 
> This is "tricky" because limits on cartesian coordinates in the real world depend on the distance. And in the unit vector they need to be calculated based on angles, which would not be very efficient. Not disastrous, but might not be worth the effort.
> 
> I believe in the particular case of double_sipm we have
> 
> ```c++
> n4::random::direction{}.axis({1, 0, 0}).max_theta(something easy to calculate).bidirectional()
> ```

But this creates a circular region where a rectangular one is desired. Not a disaster in this particular use case (some proportion of the events miss the target and are unavoidably boring), but still.

With the band stuff I imagine above, it would look like this:

``` c++
n4::random::direction{}
  .z_band_tan(full_height/distance)
  .y_band_tan(full_width /distance)
```

gonzaponte:
> The rotations would allow you to orient it in any direction.

Ah, I was not getting it right.

> The trouble with `axis` is that it's not clear how the directions orthogonal to z are re-oriented

It should be possible to determine it, but rotations should be more general. `axis` is perhaps also useful if you just want to point in one direction without thinking about angles too much.

> The G4 GUI is probably the best we can do in the sort term. We'll need some pre-built app that contains nothing but an n4::random::direction sample and pops up its result in the GUI.

Sounds good.

> With the band stuff I imagine above, it would look like this:
```
n4::random::direction{}
  .z_band_tan(full_height/distance)
  .y_band_tan(full_width /distance)
```

I like it. My only concern is how this is integrated with the rest. Ensuring that the result makes sense is difficult if not impossible.

doc/src/how-to/generate_random_numbers.md
```diff
@@ -0,0 +1,117 @@
...
+The `<optional extra specifications>` are described below. `θ` (theta) is the
+polar angle measured with respect to the positive z-axis and φ (phi) is the
+azimuthal angle with `φ=0` in the plane of the positive x-axis and increasing
+towards the positive y-axis, unless the axes are re-oriented with `.axis` or
+`.????`.
+
+The following specifiers (with hopefully self-explanatory names) restrict the angles of the generated directions
+
+- `min_theta(<angle in radians>)` TODO what happens if you pass <0
+- `max_theta(<angle in radians>)` TODO what happens if you pass >π
```
gonzaponte, line 68: I have in my branch range-checking implemented.

doc/src/how-to/generate_random_numbers.md
```diff
@@ -0,0 +1,117 @@
...
+- `min_theta_deg(<angle in degrees>)` TODO missing
+- `max_theta_deg(<angle in degrees>)` TODO missing
```
gonzaponte, line 70: Is this worth it? I don't know if there should only one way to do it, but in this ecosystem it's more natural to say `min_theta(angle_in_deg * deg)`

doc/src/how-to/generate_random_numbers.md
```diff
@@ -0,0 +1,117 @@
...
+- `min_cos_theta(<ratio>)`
+- `max_cos_theta(<ratio>)`
+
+TODO what happens if max < min
+
+- `min_phi(<angle in radians>)` TODO what is the domain and what happens if you give a value outside
+- `max_phi(<angle in radians>)` TODO what is the domain and what happens if you give a value outside
```
gonzaponte, line 77: Also in my branch

doc/src/how-to/generate_random_numbers.md
```diff
@@ -0,0 +1,117 @@
...
+  reflection in the origin.
+
+- `exclude()`: invert the selection. It generates directions that do
+  not satisfy any of the selected criteria. Bear in mind that this
+  method is more computationally expensive as it needs to generate
+  directions until the result not rejected. Very restricted
+  directionalities are discouraged.
+
+  TODO: hmm, can we not do better than rejection sampling here?
+
```
gonzaponte, line 99: In general? Maybe. Worth the complexity? Not sure.

doc/src/how-to/generate_random_numbers.md
jacg: Ah, indeed. But that possibility is easy to overlook, as I did. Still probably not worth it.

jacg:
Do we need this much generality?

Do we have some examples of this sort of thing being used in real code? double-sipm is one example, it needs a rectangular restriction. Can we find any examples to justify the complexity introduced by the others?

jacg:
Or maybe we want separate generators for 

+ phi/theta
+ rectangles via bands

and we could use a typestate to start off from the same name, and then switch into angle or band mode, as soon as the first method is used.

gonzaponte:
The idea of making bands came from double-sipm. I now realize that the generation in double-sipm is not isotropic in the restricted range given by the size of the scintillators. This is something we don't want to propagate to our band-based implementation. And now I'm struggling to understand how we can easily generate data in the intersection of two bands without a method based on rejection.

If there isn't one, perhaps it's better to provide a tool that helps reject directions. Something like `intersects(G4ThreeVector& origin, G4ThreeVector& direction, G4VSolid* target)`. What do you think?

Issue #155, opened 2023-10-16
Consider implementing an optional/maybe type


gonzaponte:
Presumably called `n4::maybe` or `n4::option` and similar to Rust's `Option`. Something like c++'s `std::optional` but without UB. It might also be worth looking into implementing a result type (`n4::result`).

Origin of discussion came from #151, where `nullptr`s can arise and must be handled.

PR #156, opened 2023-10-16
Implement `find_solid` with automatic downcast

### LINKED ISSUE #151: Implement `n4::find_solid<DOWN>` template
... for easier access to user-defined objects with automatic downcasting.

gonzaponte:
Opening this to discuss. Closes #151.

This would benefit from #155. For now, it's implemented with exceptions.

Questions:
- `static_cast` is too weak and doesn't protect against downcasting to the wrong type. I've switched to `dynamic_cast` to improve on this. Are we satisfied? In practice, this function should not be called too frequently, so this solution could be good enough.
- At the moment both checks throw the same exception. While the first one might be correct, the second one should probably throw `std::bad_cast`, which does not accept a string as an argument.
- The first of the two checks should be implemented in the non-downcasting homologous functions `find_solid`, `find_logical`, and `find_physical`


gonzaponte:
> I guess because my brain simply assumed you would know to do the right thing

I've always tried to avoid casting as much as possible, so I'm not completely familiar with the differences. Now that I had to look into it properly, I will remember it.

> So we want to return a `Result` instead of an `Option`

Yup, but the question here was more related to the exception type. I would like to throw a `std::bad_cast` exception in the second case (it seems more appropriate), but I cannot provide an error message for that exception. The best I could do was to print something before throwing the exception. I'm afraid the relevant message is lost in other noise.

> So we probably want the exception. If we manage to get usable ADTs, we can consider adding ADT-returning versions and eventually deprecating the old ones.

I would certainly move on with an exception while we consider ADTs. That change will likely impact other code as well.

jacg:
> the question here was more related to the exception type.

+ Throw our own exception type?
+ Use the G4 exception stuff? (I know very little about this.)

gonzaponte:
> Use the G4 exception stuff? (I know very little about this.)

Same. I'm not very interested in delving into G4Exceptions. Would you be happy with 

>Throw our own exception type

?

gonzaponte:
Please take a look at the last commit and let me know if you have a preference.

nain4/src/meson.build
```diff
@@ -37,6 +37,7 @@ nain4_includes = [ 'n4-all.hh'
                  , 'n4-boolean-shape.hh'
                  , 'n4-constants.hh'
                  , 'n4-defaults.hh'
+                 , 'n4-exceptions.hh'
```
jacg, line 40: Mustn't forget to add it to the mdbook.

jacg:
If you're doing more than the absolute bare minimum on these exceptions, then we probably *should* look at the G4 exception stuff.

jacg:
I prefer the intermediate `n4_exception`. Perhaps it should be called `nain4::exception`.

jacg:
Probably because G4 catches G4 exceptions (and then aborts in some cases)? I'm struggling to find the spare brain cycles to care about G4 exceptions, but I'm slightly uneasy about overlooking something important, by ignoring them.

gonzaponte:
I've dug into the source code a bit and didn't find any relevant throw clauses on the kernel side. Looks like they simply call `abort()` after printing different messages.

nain4/src/n4-utils.hh
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <n4-constants.hh>
+#include <n4-exceptions.hh>
```
jacg, line 4: I'm not convinced that `<n4-exceptions.hh>` belongs in `<n4-utils.hh>`.

I think that very few clients will ever want to throw or catch one of these, so they have no use for the headers whatsoever.

I'd add it to `<n4-all.hh>` (if that) and nowhere else.

PR #190, opened 2024-11-07
Avoid double geant4 compilation in flake

gonzaponte:
Our flake was using the default geant4 in the devshell and our geant4 everywhere else, forcing nix to compile geant4 twice. The only differences between the two outputs are the overridden attributes. I think this solves the problem, however, it's difficult to test whether this is true once the dependencies are in the nix store. Do you have any hints on how to test this?


gonzaponte:
After `nix build`, `nix-store --query --requisites result | rg geant` yields
```
/nix/store/42zl2p52lhxffbqr8qa0c55h8ql76m6w-geant4-11.2.1
/nix/store/9ixk9qjj2qwyg5nzfxxvjb8ppw55p1cr-geant4-11.2.1
```

so it seems the issue is not solved just yet.

jacg:
Don't forget that `nix build ...` puts packages (default or otherwise) into `result`, so you haven't discovered anything about the devshell yet: the problem seems to be at a lower level than specifying multiple G4s in the shell's definition itself.

gonzaponte:
For some reason, now I'm getting a previous version of geant4:

```
/nix/store/nwfmazzgpa78yjk87fb8crd8v3y13b1v-geant4-11.0.4
```

But we only get one geant4 instance, as expected.

PR #203, opened 2025-12-10
Add action builder

gonzaponte:
A partial implementation of #30. The idea is simple: create a more comfortable interface for setting actions. A complete configuration would look like:

```
n4::action_builder{a_generator}
  .run_begin          (...)
  .run_end            (...)
  .event_begin        (...)
  .event_end          (...)
  .stacking_classify  (...)
  .stacking_next_stage(...)
  .stacking_next_event(...)
  .tracking_pre       (...)
  .tracking_post      (...)
  .stepping           (...)
  .done               ();
```
But all of these actions are optional.

In this implementation, it was quite natural to include another perk: the possibility of combining multiple actions of the same kind. In my experience, this can be quite useful. It allows the user to create small, well-defined functions that do something very specific, very clearly. Giving these functions meaningful names can result in a self-explanatory workflow. For instance:
```
n4::action_builder{isotropic_2614keV_gammas_from_origin}
  .event_begin (print_event_number)
  .event_begin (initialize_something)
  .tracking_pre(count_compton)
  .tracking_pre(count_pair_production)
  .tracking_pre(count_photoelectric)
  .tracking_pre(kill_positrons_because_they_are_too_positive)
  .etc
``` 

The current implementation does not allow for classes to be passed to this builder, but I believe it is feasible to implement a version that does. In any case, I first wanted to hear your thoughts (@jacg) on the current approach, if you have any.


jacg:
You've caught me in a bad week ... or two. I won't be able to give this the attention it deserves for a while. Is it urgent?

gonzaponte:
Not at all.

